PR #98, opened 2017-11-09
Add Jest and Enzyme with beginning front end test

### LINKED ISSUE #80: As a developer, I want a test framework for javascript frontend and backend
Acceptance Criteria:

- [ ] report with code coverage for both frontend and backend
- [ ] test: frontend, the homepage renders
- [x] test: backend, 404s on unknown route, is `application/json`

chtinahow:
Resolves #80 
There is an issue with the babelrc file where it's longer than it should be. Maybe because of Dylan?
I can't get the tests to run properly with that file and I'd like help figuring that out before this can be merged in.


billdybas:
Frontend testing is now setup. Also includes a coverage report.

If you run this branch locally, the coverage report will yell at you about [this](https://github.com/robmcl4/A-B-S-T-R-A-C-T-I-O-N/blob/fcda76b0619ecee6c84708d5192266a373bdc9f6/frontend/src/Admin/reducers.js#L5) line which I've fixed in another branch, so ignore it for now.

PR #105, opened 2017-11-15
Adding judges query tests

kayladavis:


backend/test/resolvers/userQuery.js
```diff
@@ -0,0 +1,101 @@
...
+      fakeUser({username: 'user1'}).then((u) => {
+        user('', {id: u.username}, {auth: {type: 'STUDENT', id: 'user1'}})
+          .then((result) => {
+            expect(result.username).to.equal(u.username)
+            done()
+          })
+      })
+    })
+    it('Does not allow non admins to find other users by username', function (done) {
+      fakeUser({username: 'user1'}).then((u) => {
```
chtinahow: What is u in this case?

kayladavis: user, it's a u because if I write 'user' then the function I'm testing can't run. Do you want me to change it.

kayladavis:
I made the description capitalization changes. The travis fails is because rob hasn't figured out how to hook it up yet.

PR #107, opened 2017-11-18
Add query resolver for an entry

robmcl4:
Adds a query resolver for an entry. This is accomplished by:

* Adding `Group` model with attributes `id`, and `name`
* Adding `Image` model with attributes `id`, `path`, `horizDimInch`, `vertDimInch`, and `mediaType`
* Adding `Entry` model with a whole lot of attributes (see code)
* Adding a `fakeImageEntry(opts)` method for testing

Also included, major test speadup by changing how models are cleaned up after use.

Suite execution time decreated from 40s to 2s (!!!)

### DIFF
--- a/backend/constants.js
+++ b/backend/constants.js
@@ -1,0 +1,6 @@
+export const STUDENT = 'STUDENT'
+export const ADMIN = 'ADMIN'
+export const JUDGE = 'JUDGE'
+export const IMAGE_ENTRY = 0x1
+export const VIDEO_ENTRY = 0x2
+export const OTHER_ENTRY = 0x3
--- a/backend/db/migrations/20171116200117-create-group.js
+++ b/backend/db/migrations/20171116200117-create-group.js
@@ -1,10 +1,14 @@
 export function up (queryInterface, Sequelize) {
-  return queryInterface.createTable('group', {
+  return queryInterface.createTable('groups', {
     id: {
       allowNull: false,
       autoIncrement: true,
       primaryKey: true,
       type: Sequelize.INTEGER
+    },
+    name: {
+      allowNull: false,
+      type: Sequelize.STRING
     },
     createdAt: {
       allowNull: false,
--- a/backend/db/migrations/20171116292217-create-entry.js
+++ b/backend/db/migrations/20171116292217-create-entry.js
@@ -71,6 +71,10 @@ export function up (queryInterface, Sequelize) {
       allowNull: true,
       type: Sequelize.TEXT
     },
+    excludeFromJudging: {
+      allowNull: false,
+      type: Sequelize.BOOLEAN
+    },
     createdAt: {
       allowNull: false,
       type: Sequelize.DATE
--- a/backend/db/migrations/20171116292217-create-entry.js
+++ b/backend/db/migrations/20171116292217-create-entry.js
@@ -63,6 +63,14 @@ export function up (queryInterface, Sequelize) {
       allowNull: true,
       type: Sequelize.BOOLEAN
     },
+    yearLevel: {
+      allowNull: true,
+      type: Sequelize.TEXT
+    },
+    academicProgram: {
+      allowNull: true,
+      type: Sequelize.TEXT
+    },
     createdAt: {
       allowNull: false,
       type: Sequelize.DATE
--- a/backend/db/migrations/20171116292217-create-entry.js
+++ b/backend/db/migrations/20171116292217-create-entry.js
@@ -6,13 +6,32 @@ export function up (queryInterface, Sequelize) {
       primaryKey: true,
       type: Sequelize.INTEGER
     },
-    entrantType: {
-      allowNull: false,
-      type: Sequelize.INTEGER
+    showId: {
+      type: Sequelize.INTEGER,
+      references: {
+        model: 'shows',
+        key: 'id'
+      },
+      onUpdate: 'cascade',
+      onDelete: 'cascade'
     },
-    entrantId: {
-      allowNull: false,
-      type: Sequelize.INTEGER
+    studentUsername: {
+      type: Sequelize.STRING,
+      references: {
+        model: 'users',
+        key: 'username'
+      },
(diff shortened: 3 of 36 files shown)


backend/models/entry.js
```diff
@@ -0,0 +1,79 @@
+import Image from './image'
+import DataTypes from 'sequelize'
+import sequelize from '../config/sequelize'
+
+const IMAGE = 3
```
chtinahow: Where are the magic numbers coming from?

backend/db/migrations/20171116292217-create-entry.js
```diff
@@ -0,0 +1,79 @@
...
+    showId: {
+      type: Sequelize.INTEGER,
+      references: {
+        model: 'shows',
+        key: 'id'
+      },
+      onUpdate: 'cascade',
+      onDelete: 'cascade'
+    },
+    studentUsername: {
```
kayladavis: I think this follows your erd diagram, not Bill's. Did we decide to use your approach instead? Or was it just much harder to implement his his method?

backend/models/entry.js
kayladavis: Is it worth it to make these enums and move them out to another file?

kayladavis:
please add year level and program to entry type

robmcl4:
plz change `permissionLevels.js` to `constants.js` and do things to move magic numbers there

backend/models/entry.js
robmcl4: They're just a mimic-enum -- i'll move them to a `constants.js` file

backend/db/migrations/20171116292217-create-entry.js
robmcl4: Unfortunately `entityId` can't be both type INT (to support groups) and STRING (to support users / usernames).

backend/models/entry.js
robmcl4: Will do

backend/constants.js
```diff
@@ -0,0 +1,6 @@
+export const STUDENT = 'STUDENT'
```
chtinahow: Yaay constants file!

backend/resolvers/types/entryType.js
```diff
@@ -0,0 +1,16 @@
+import { IMAGE_ENTRY, VIDEO_ENTRY, OTHER_ENTRY } from '../../constants'
+
+export default {
+  Entry: {
+    __resolveType (data, context, info) {
+      if (data.entryType === IMAGE_ENTRY) {
```
chtinahow: I think this could be written more elegantly, but it wouldn't matter too much.

backend/test/factories.js
```diff
@@ -31,13 +33,77 @@ export function fakeShow (opts) {
...
+        showId: show.id,
+        studentUsername: user ? user.username : null,
+        groupId: opts.group ? opts.group.id : null,
+        entryType: entryType,
+        entryId: entryId,
+        title: opts.title,
+        comment: opts.comment,
+        moreCopies: opts.moreCopies,
+        forSale: opts.forSale,
+        awardWon: opts.awardWon,
```
chtinahow: awardWon?

backend/resolvers/types/entryType.js
robmcl4: agreed, but I dunno ... we can make it better if we think of what to do. For background, this is a magic method for GraphQL so it knows what type of the abstract `Entry` interface you're returning, so it extracts the right properties

kayladavis: I don't think we need to change this now. It's pretty clear as is. If we need to add to it later, let's make sure it's clear.

backend/test/factories.js
robmcl4: it's documented in Bill's version of the ERD: iirc, during the gallery there's some various named awards that are issued to individual works, and she documents them in the system. We should ask if she wants that feature still. The field is otherwise optional so... we can leave it in and not use it if needed

backend/db/migrations/20171116292217-create-entry.js
```diff
@@ -0,0 +1,91 @@
...
+      allowNull: true,
+      type: Sequelize.TEXT
+    },
+    academicProgram: {
+      allowNull: true,
+      type: Sequelize.TEXT
+    },
+    excludeFromJudging: {
+      allowNull: false,
+      type: Sequelize.BOOLEAN
```
billdybas: I would add `defaultValue: false`

backend/db/migrations/20171116200117-create-group.js
```diff
@@ -0,0 +1,26 @@
...
+      allowNull: false,
+      autoIncrement: true,
+      primaryKey: true,
+      type: Sequelize.INTEGER
+    },
+    name: {
+      allowNull: false,
+      type: Sequelize.STRING
+    },
+    createdAt: {
```
billdybas: I would add `defaultValue: Sequelize.fn('NOW')` to all the `createdAt` and `updatedAt` fields

backend/db/migrations/20171116292217-create-entry.js
```diff
@@ -0,0 +1,91 @@
...
+    forSale: {
+      allowNull: false,
+      type: Sequelize.BOOLEAN
+    },
+    awardWon: {
+      type: Sequelize.TEXT
+    },
+    invited: {
+      allowNull: true,
+      type: Sequelize.BOOLEAN
```
billdybas: I would add `defaultValue: false`

backend/models/entry.js
```diff
@@ -0,0 +1,94 @@
...
+        throw new Error('Entry must have an entrant')
+      }
+    }
+  }
+})
+
+/*
+ * Gets the associated photo as a Promise
+ */
+Entry.prototype.getImage = function getImage () {
```
billdybas: Could this be added in `instanceMethods` or does it have to be added to the prototype like this?

PR #109, opened 2017-11-23
Assigning Judges to a Show Form

billdybas:
Adds a page where Admins can assign Judges to shows.

Also adds seeds for Users on the backend.

~~~Going forward, we're going to switch out reactstrap for [react-semantic-ui](https://github.com/Semantic-Org/Semantic-UI-React) as reactstrap has been a pain to work with and doesn't seem to conform to React conventions.~~~ [Edit: Not worth the effort to switch out, and quick tests show that semantic-ui's stylesheet has problems with the date picker's styles we're using.]

~~~Still working on refetching data & updating the store after an assignment is made. Also want to pull in https://github.com/gorangajic/react-icons for Font Awesome Icons (currently using Open Iconic).~~~ [Edit: ✅]

~~~Will remove WIP status when things are finished. Review at your leisure, but code will change.~~~ [Edit: ✅]


frontend/src/Admin/components/AssignJudgesTable.js
```diff
@@ -0,0 +1,138 @@
...
+  handleAssignedJudgesChange = (selectedAssignedJudges) => {
+    this.setState({selectedAssignedJudges})
+  }
+
+  handleUnassignedJudgesChange = (selectedUnassignedJudges) => {
+    this.setState({selectedUnassignedJudges})
+  }
+
+  render () {
+    const {
```
chtinahow: Just const?

frontend/src/Admin/containers/AssignJudgesTable.js
```diff
@@ -0,0 +1,84 @@
+import { graphql, compose } from 'react-apollo'
+import { connect } from 'react-redux'
+
+import {
+  feetchJudgesByAssignmentForShow,
```
chtinahow: feetch or fetch?

frontend/src/Admin/components/AssignJudgesTable.js
billdybas: This is destructuring `data` from `this.props`, so no.

frontend/src/Admin/containers/AssignJudgesTable.js
billdybas: Good catch. Copied and pasted this from `actions.js` which is also spelled wrong, but since they're the same, it doesn't cause issues.

frontend/config/apollo.js
```diff
@@ -18,7 +18,17 @@ const authMiddleware = new ApolloLink((operation, forward) => {
 
 const client = new ApolloClient({
   link: concat(authMiddleware, httpLink),
-  cache: new InMemoryCache()
+  cache: new InMemoryCache({
+    dataIdFromObject: o => {
+      switch (o.__typename) {
+        case 'User':
+          // The primary key of 'User' is 'username'
+          return o.username
```
kayladavis: If using username as the primary key proves to be such an issue we can take the time to change it back to the id. We talked offline about this being easier for frontend, but it looks like it's possibly causing issues. @robmcl4 for his feelings about this.

frontend/config/apollo.js
```diff
@@ -18,7 +18,17 @@ const authMiddleware = new ApolloLink((operation, forward) => {
 
 const client = new ApolloClient({
   link: concat(authMiddleware, httpLink),
-  cache: new InMemoryCache()
+  cache: new InMemoryCache({
+    dataIdFromObject: o => {
+      switch (o.__typename) {
+        case 'User':
+          // The primary key of 'User' is 'username'
```
kayladavis: Is is worth it to go back to username's pk equaling an Id? We talked about usernames making more sense weeks ago, but it seems to be causing problems.

frontend/src/Admin/actions.js
```diff
@@ -0,0 +1,84 @@
...
+    query: JudgesForShowQuery,
+    variables: {
+      id: showId
+    }
+  })
+    .then(({data: {show}}) => dispatch({type: FETCH_JUDGES_FOR_SHOW, payload: show}))
+    .catch(console.error) // TODO: Handle the error
+}
+
+export const feetchJudgesByAssignmentForShow = (showId) => (dispatch, getState, client) => {
```
kayladavis: adding to Tina's point. 'feetch' should be fetch

frontend/src/Admin/pages/AssignJudges.js
```diff
@@ -0,0 +1,18 @@
...
+import AssignJudgesTable from '../containers/AssignJudgesTable'
+
+const Heading = styled.h1`
+  margin-bottom: 25px;
+`
+
+const ManageJudges = (props) => (
+  <Container fluid>
+    <Heading>Manage Judges</Heading> {/* TODO: Mention the name of the show */}
+    <AssignJudgesTable showId={props.match.params.id} /> {/* TODO: Handle non-number input more gracefully */}
```
kayladavis: Can you explain this todo? The show id is an int, what's wrong here?

frontend/config/apollo.js
billdybas: If it's not too much effort to switch to id's, we could, as Apollo automatically supports objects w/ an `id` or `_id` field; however, I don't think it's any more difficult on the frontend w/ usernames. This config here allows Apollo's store to keep track of how different User update from mutations.

backend/config/database.js
```diff
@@ -5,27 +5,29 @@ const HOST = process.env.ABSTRACTION_DB_HOST
...
-  port: 3306
+  database: 'test',
+  username: USER,
+  password: PASSWORD,
+  port: 3306,
+  logging: false
 }
 
-export default {
+// Additonal exports so that sequelize-cli can read this config properly
```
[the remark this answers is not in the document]
billdybas: `export default {}` is essentially syntactic sugar for `module.exports.default = {}`

You can still import this file like we do in others: `import dbConfig from 'config/database'`
That same statement in CommonJS format is `const dbConfig = require('config/database').default`

Because `sequelize-cli` reads in this file directly, it's expecting the modules to be exported using CommonJS format (`module.exports`). We might be able to hook Babel into the `sequelize-cli` process, but that seems unnecessary since this code already works.

frontend/src/Admin/pages/AssignJudges.js
billdybas: You are correct - the show id will always work because it is an integer, however, `react-router` doesn't limit the type of `id`, so technically you can type into your browser `/admin/show/a/judges` and you would hit the correct page, but Apollo blows up because it doesn't know how to interpret `a` and an id.

I think I have an idea how to fix this by showing the Not Found page if it gets and id it doesn't expect.

frontend/src/Admin/containers/AssignJudgesTable.js
```diff
@@ -0,0 +1,84 @@
...
+
+const mapDispatchToProps = (dispatch, {showId}) => ({
+  fetchData: () => dispatch(fetchJudgesByAssignmentForShow(showId)),
+  afterAssign: (usernames) => dispatch(assignJudgesToShow(showId, usernames)),
+  afterUnassign: (usernames) => dispatch(removeJudgesFromShow(showId, usernames))
+})
+
+const withRedux = connect(mapStateToProps, mapDispatchToProps)(AssignJudgesTable)
+
+const withMutations = compose(
```
robmcl4: I feel your pain, this is :(

frontend/src/Admin/actions.js
```diff
@@ -0,0 +1,97 @@
...
+      client.query({
+        query: JudgesForShowQuery,
+        variables: {
+          id: showId
+        }
+      }),
+      dispatch(fetchShows()),
+      dispatch(fetchJudges())
+    ])
+    .spread(({data: {show}, loading}) => {
```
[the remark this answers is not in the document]
billdybas: Based on the way we normalize our state and how Apollo caches data, if someone goes to the Judge assignment page directly, we need to load in Shows, Judges, and the Judge assignments between shows, which is why I dispatch 3 actions here. 

To answer 1, I need to look into the React lifecycle w/ Apollo more because uninitialized data (eg. `null`, `undefined`) seems to flow through `render` a few times before actual data. Now that I re-look at this, I don't even know if `loading` will be false based on using `Promise.all`. I'll have to look into Apollo's `loading` (and `error`) props more because their docs only have examples using their `graphql` higher-order function, not w/ using Redux.

To answer 2, pretty sure it does. See `mapDispatchToProps` in one of the containers below. Has to do with how `fetchShows()` is a function which returns a function.

frontend/src/Admin/actions.js
```diff
@@ -0,0 +1,97 @@
...
+export const FETCH_SHOW = 'FETCH_SHOW'
+export const FETCH_SHOWS = 'FETCH_SHOWS'
+export const FETCH_JUDGES = 'FETCH_JUDGES'
+export const FETCH_JUDGES_FOR_SHOW = 'FETCH_JUDGES_FOR_SHOW'
+export const FETCH_JUDGES_BY_ASSIGNMENT_FOR_SHOW = 'FETCH_JUDGES_BY_ASSIGNMENT_FOR_SHOW'
+export const ASSIGN_JUDGES_TO_SHOW = 'ASSIGN_JUDGES_TO_SHOW'
+export const REMOVE_JUDGES_FROM_SHOW = 'REMOVE_JUDGES_FROM_SHOW'
+
+export const fetchShow = (showId) => (dispatch, getState, client) => {
+  return client.query({
```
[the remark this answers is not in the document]
billdybas: non-null on what?

We have to pass `showId` into this function and if `showId` isn't valid, then we'll get a GraphQL error back which will be caught.

`client` will always be defined as it's passed in through the store (See: `thunk.withExtraArgument(client)`)

frontend/src/Admin/components/AssignJudgesTable.js
```diff
@@ -0,0 +1,138 @@
...
+  assign = () => {
+    const judges = Object.keys(this.state.selectedUnassignedJudges)
+
+    if (judges.length) {
+      this.props.assign(judges)
+        .then(() => {
+          this.props.afterAssign(judges)
+          // Reset the checkboxes
+          this.setState({
+            selectedUnassignedJudges: {},
```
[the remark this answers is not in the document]
billdybas: Why do you ask?

The reason I do this is because data for both tables is refetched after assigning/unassigning Judges, so to keep both tables selection state in sync, I clear them after an action. The alternative is to do some weird merging / removing of incoming and outgoing data in each tables which is not worth the effort. Clearing the selections feels to have decently good UX.

frontend/src/Admin/components/AssignJudgesTable.js
```diff
@@ -0,0 +1,138 @@
...
+    const judges = Object.keys(this.state.selectedAssignedJudges)
+
+    if (judges.length) {
+      this.props.unassign(judges)
+        .then(() => {
+          this.props.afterUnassign(judges)
+          // Reset the checkboxes
+          this.setState({
+            selectedUnassignedJudges: {},
+            selectedAssignedJudges: {}
```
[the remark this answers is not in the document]
billdybas: See above comment.

billdybas:
@robmcl4 I'm going to merge this now so that you can work on the Add Judge Button which goes on one of the pages in this PR. I was planning on adding reducer tests, but they're less of a priority right now and can wait for a future PR.

PR #110, opened 2017-11-27
Adding an image uploading endpoint

kayladavis:
An uploading endpoint that uses `multer` to handle file uploads.

This PR is to make sure that this endpoint makes sense for the ideas that Bill/Tina have about image uploading.

### DIFF
--- a/backend/server.js
+++ b/backend/server.js
@@ -29,6 +29,6 @@ app.use('/graphql', graphqlHttp(req => ({
   context: req
 })))
 
-router.post('/static/upload', imageUploader)
+router.post('/static/upload', imageUploader) // TODO: Require Authentication
 
 export default app // for testing
--- a/backend/server.js
+++ b/backend/server.js
@@ -10,6 +10,7 @@ import models from './models'
 import router from './routes'
 import schema from './schema'
 import parseJwtUser from './middleware/parseJwtUser'
+import imageUploader from './upload'
 
 const app = express()
 models()
@@ -27,5 +28,7 @@ app.use('/graphql', graphqlHttp(req => ({
   graphiql: config.get('NODE_ENV') !== 'production',
   context: req
 })))
+
+router.post('/upload', imageUploader)
 
 export default app // for testing
--- a/backend/server.js
+++ b/backend/server.js
@@ -29,6 +29,6 @@ app.use('/graphql', graphqlHttp(req => ({
   context: req
 })))
 
-router.post('/upload', imageUploader)
+router.post('/static/upload', imageUploader)
 
 export default app // for testing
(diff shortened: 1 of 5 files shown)


backend/upload.js
```diff
@@ -0,0 +1,24 @@
+import multer from 'multer'
+
+var Storage = multer.diskStorage({
```
billdybas: Convert to ES6 throughout this file (eg. `const`, `(err) => {}`)

backend/upload.js
```diff
@@ -0,0 +1,24 @@
...
+    callback(null, './images')
+  },
+  filename: function (req, file, callback) {
+    callback(null, file.fieldname + '_' + Date.now() + '_' + file.originalname)
+  }
+})
+
+var upload = multer({
+  storage: Storage,
+  limits: {fileSize: 1000000, files: 1} // Max 1 MB
```
billdybas: 1MB is likely too low a limit. 50MB feels like a good number to me since highly-edited, hi-res images can be quite large. We should talk to Nanette about typical past file sizes.

backend/upload.js
```diff
@@ -0,0 +1,24 @@
...
+})
+
+export default function uploader (req, res) {
+  upload(req, res, function (err) {
+    if (err) {
+      return res.end('Something went wrong!')
+    }
+    return res.end('File uploaded sucessfully!.')
```
billdybas: I would send a JSON response back w/ the path of where the file was uploaded.

backend/upload.js
```diff
@@ -0,0 +1,24 @@
+import multer from 'multer'
+
+var Storage = multer.diskStorage({
+  destination: function (req, file, callback) {
+    callback(null, './images')
```
billdybas: We need more information from the Sysadmins where we're actually going to be storing the images (i.e. if they're going on the same hard-drive as our system or on 3rd-party object storage).

Also, the convention I've seen for directory / file naming is to generate a unique hash and save the file based on that hash. For example:
Uploaded Image: `my_awesome_photo.jpg`
Could be hashed to: `kTK6vi6Z4IAlh65eSMps.jpg`
And would be saved in the directory hierarchy: `k/T/kTK6vi6Z4IAlh65eSMps.jpg`

We'd rename the files when they're downloaded (based on the student's name and submission title). If we needed to keep the original file name, we could store it in the db or append it to the hash (e.g. `kTK6vi6Z4IAlh65eSMps_my_awesome_photo.jpg`)

backend/server.js
```diff
@@ -28,4 +29,6 @@ app.use('/graphql', graphqlHttp(req => ({
   context: req
 })))
 
+router.post('/upload', imageUploader)
```
billdybas: We might want endpoints to fetch images and/or download images, so I'd namespace this route to `/static/upload`.

PR #113, opened 2017-12-07
Adding video model, 'create' mutation, and query 

kayladavis:


billdybas:
I quickly glanced at the video regex, and I don't think it covers all cases. 

You can find more robust regexes and some examples here:
- YouTube:
https://github.com/CookPete/react-player/blob/master/src/players/YouTube.js#L8
https://github.com/CookPete/react-player/blob/master/test/specs/canPlay.js#L14-L18 (These probably also need more query parameter tests (eg. when `?v=` is `&v=`))
- Vimeo: 
https://github.com/CookPete/react-player/blob/master/src/players/Vimeo.js#L7
https://github.com/CookPete/react-player/blob/master/test/specs/canPlay.js#L43-L50

backend/constants.js
```diff
@@ -4,3 +4,4 @@ export const JUDGE = 'JUDGE'
 export const IMAGE_ENTRY = 0x1
 export const VIDEO_ENTRY = 0x2
 export const OTHER_ENTRY = 0x3
+export const ALLOWED_VIDEO_PROVIDERS = new Set(['youtube', 'vimeo'])
```
robmcl4: Is there an immutable set?

backend/schema.js
```diff
@@ -79,7 +79,6 @@ interface Entry {
     yearLevel: String
     academicProgram: String
     moreCopies: Boolean
-    excludeFromJudging: Boolean
```
robmcl4: ty

backend/constants.js
kayladavis: apparently you can't do this w/o a dependency so it's staying as is.

Issue #15, opened 2017-09-19
As a student, I would like to be able to drag-and-drop the files I am uploading to my submission, so that I don’t have to navigate a file browser in my web browser for them if I already have them ready


robmcl4:
As a student, I would like to be able to drag-and-drop the files I am uploading to my submission, so that I don’t have to navigate a file browser in my web browser for them if I already have them ready

billdybas:
https://react-dropzone.js.org/

PR #115, opened 2017-12-08
Student Image Entry Dropzone

billdybas:
This PR adds partial completion of the Student entry submission form.

#15 only specifies a dropzone component which is now implemented on this page - uploading an image saves it to the backend and renders a preview of that image on the client (inside the dropzone) + redux form knows about the `path` value. Therefore, I'm considering this story done, and just made another story for the actual submission form.

I also started adding additional form fields + redux form + saving to backend, but am blocked by 2 things:
1. I need further clarification on how to word some questions (eg. `moreCopies`)
2. The GraphQL schema for `EntryInput` does not match what is implemented - I get Sequelize errors when trying to create an image entry based on it (eg. `moreCopies` is listed as not required, but I get not null constraint errors)


billdybas:
Merging this in. Currently inaccessible until #117 is merged in, but this form still needs a lot of work until feature-complete. 

@kayladavis @robmcl4 See my original comment, point 2, as this feature will be blocked by backend issues.

PR #116, opened 2018-01-18
npm bootstrap type error fix

kayladavis:


robmcl4:
Pulled, confirmed that this fixes the problem. LGTM.

PR #117, opened 2018-01-18
 Redirect logged-in users to correct routes

robmcl4:
Changes the login flow so that users are redirected to either `/admin/`, `/judge/`, or `/student/` depending on their user type. This is accomplished by:

1) changing the backend so that post-login also sets a `_user_type_v1` cookie, in addition to the existing `_token_v1` cookie.
2) Changing the frontend so that on boot it searches for this `_user_type_v1` cookie, stores it in localstorage, and deletes the cookie. The `Home` component searches for these localstorage items and redirects to each user's page, appropriately.

Builds on top of #116


billdybas:
Here's an idea. Right now if we include /student, /judge, /admin, a crafty user could guess the url of the other users' pages. Therefore we'd have to add in display logic to not show things if a user's not authorized or redirect them to the pages they have permission to view. The API token does have auth checking so sensitive info wouldn't leak. 

However, what if we get rid of the URL based name spacing, and simply load the app the user is authorized to use. Therefore, if a student goes to /dashboard, they see a dashboard meant for them, but an admin who goes to /dashboard would see their admin dashboard. 

We could keep the user's type in local storage like you already have, but use react-router's Switch component to render the appropriate pages based on the user's type.

billdybas:
I made some edits. See my commit messages for more details. Only _major_ change was removing the token when you log out – prior, if you refreshed the page after logging out, you'd immediately be logged in.

LGTM 🚀

PR #123, opened 2018-01-21
Backend Updates

billdybas:
Cleans up some backend code, switches `express-graphql` for `apollo-server-graphql`, and adds in a more robust video URL parsing regex.


backend/test/helpers/submission.js
```diff
@@ -1,18 +1,20 @@
-import { parseVideo } from '../../helpers/submittion'
 import { expect } from 'chai'
 
+import { parseVideo } from '../../helpers/submission'
+
 describe('Parse Video URL', function () {
-  it('accepts valid youtube urls', function () {
```
robmcl4: we're not really doing anything asynchronous here, why have a `done`?

billdybas: Good catch! There was an unneeded `return Promise.resolve()` there before and my brain subconsciously changed it to `done()`. It's now removed.

PR #124, opened 2018-01-21
Robert/add other media

robmcl4:
Builds ontop of #123. Adds an 'Other Media' mutation / query, including tests.

OtherMedia is an Entry which has these unique properties:
* `path` - a path to a photo, non-null


robmcl4:
Er... video submissions were added in a previous PR. We have three right now, "image", "video" and "other"

backend/db/migrations/20180121163800-create-other.js
```diff
@@ -0,0 +1,25 @@
...
+      autoIncrement: true,
+      primaryKey: true,
+      type: Sequelize.INTEGER
+    },
+    path: {
+      type: Sequelize.STRING
+    },
+    createdAt: {
+      allowNull: false,
+      type: Sequelize.DATE
```
billdybas: You should add `defaultValue: Sequelize.fn('NOW')` to this and `updatedAt`

backend/resolvers/mutations/entry.js
```diff
@@ -44,3 +45,19 @@ export function createVideo (_, args, req) {
     }).then((entry) => Object.assign(entry, video.dataValues))
   })
 }
+
+export function createOtherMedia (_, args, req) {
+  if (req.auth.type !== ADMIN && !allowedToSubmit(args, req)) {
+    // don't allow non-admins to submit work claiming to be from someone else
+    throw new UserError('Permission Denied')
+  }
+  return Other.create({
```
billdybas: Should probably be in a transaction.

billdybas:
Also, probably need to rebase off of master.

PR #125, opened 2018-01-22
Adding image upload tests and image resources

kayladavis:
- verifies images of 50+ mb can't be added
- verifies image is a jpg
- verifies one image can be sent per request
- checks persistence

I got a warning while pushing over the 58mb file. Thoughts on keeping it how we have it or changing it to the recommendation?
large_image_problem


kayladavis:
image deletion is still a problem. I can either try to delete the folder (we may run into permissions issues) or start using mocking/spys (not truly testing persistence).

backend/upload.js
```diff
@@ -2,35 +2,56 @@ import multer from 'multer'
 import mkdirp from 'mkdirp'
 import uuidv1 from 'uuid/v1'
 
+const test_dir = './test/images'
+const image_dir = './images'
+
 const Storage = multer.diskStorage({
   destination: (req, file, callback) => {
-    callback(null, './images') // TODO: See if we want a ./year/images
+    if (process.env.NODE_ENV === 'test'){
```
robmcl4: I think we should stick this into the config system.

backend/test/images/.gitignore
```diff
@@ -0,0 +1,2 @@
+*
+!.gitignore
```
robmcl4: What's this do? Would a simple `.gitkeep` suffice with a corresponding entry in `backend/.gitignore`? fine either way

backend/upload.js
```diff
@@ -2,35 +2,56 @@ import multer from 'multer'
...
+      callback(null, test_dir) // TODO: See if we want a ./year/images
+    } else {
+      callback(null, image_dir) // TODO: See if we want a ./year/images
+    }
   },
   filename: (req, file, callback) => {
     // Generate a unique id for the file
     const filename = uuidv1()
```
robmcl4: Now that I see this ... should we be using v1 here? Is there any reason to avoid v4? Hmm, will discuss later, not a problem for now

backend/test/images/.gitignore
kayladavis: We can actually probably just do with the git ignore all images. I just pulled that in from what I had in the /images.

PR #127, opened 2018-01-23
Adds new Group schema for freeform participants

robmcl4:
Changes the Group schema to allow participants to be added via a freeform text field.

Also modifies the submission mechanism to support automagically creating a group when submitting work.

Example mutation for submitting with a group:

```graphql
mutation {
  createPhoto(input:{
    entry: {
      group: {
        name:"my fun group",
        creatorUsername:"user1",
        participants:"uncle timmy"
      },
      showId: 1,
      title: "mytitle",
      comment: "this is my comment",
      forSale: false,
      yearLevel: "second",
      academicProgram: "learning",
      moreCopies: false
    },
    path:"foo.jpg",
    horizDimInch:3,
    vertDimInch:3,
    mediaType:"mymedia"
  }) {
    id
  }
}
```

### DIFF
--- a/backend/db/migrations/20171116200117-create-group.js
+++ b/backend/db/migrations/20171116200117-create-group.js
@@ -22,5 +22,5 @@ export function up (queryInterface, Sequelize) {
 }
 
 export function down (queryInterface) {
-  return queryInterface.dropTable('group')
+  return queryInterface.dropTable('groups')
 }
--- a/backend/models/image.js
+++ b/backend/models/image.js
@@ -21,6 +21,14 @@ export default sequelize.define('image', {
     type: DataTypes.STRING,
     allowNull: false,
     notEmpty: true
+  },
+  createdAt: {
+    allowNull: false,
+    type: DataTypes.DATE
+  },
+  updatedAt: {
+    allowNull: false,
+    type: DataTypes.DATE
   }
 },
 {
--- a/backend/models/other.js
+++ b/backend/models/other.js
@@ -7,6 +7,14 @@ export default sequelize.define('other', {
   path: {
     type: DataTypes.STRING,
     allowNull: true
+  },
+  createdAt: {
+    allowNull: false,
+    type: DataTypes.DATE
+  },
+  updatedAt: {
+    allowNull: false,
+    type: DataTypes.DATE
   }
 },
 {
--- a/backend/models/video.js
+++ b/backend/models/video.js
@@ -13,6 +13,14 @@ export default sequelize.define('video', {
     type: DataTypes.STRING,
     allowNull: false,
     notEmpty: true
+  },
+  createdAt: {
+    allowNull: false,
+    type: DataTypes.DATE
+  },
+  updatedAt: {
+    allowNull: false,
+    type: DataTypes.DATE
   }
 },
 {
--- a/backend/resolvers/mutations/entry.js
+++ b/backend/resolvers/mutations/entry.js
@@ -5,8 +5,34 @@ import Entry from '../../models/entry'
 import Image from '../../models/image'
 import Video from '../../models/video'
 import Other from '../../models/other'
-import { ADMIN, IMAGE_ENTRY, OTHER_ENTRY } from '../../constants'
+import Group from '../../models/group'
+import { ADMIN, IMAGE_ENTRY, OTHER_ENTRY, VIDEO_ENTRY } from '../../constants'
 import { allowedToSubmit, parseVideo } from '../../helpers/submission'
+
+// Creates an Entry based on the 'EntryInput' schema
+const createEntry = (entry, entryType, entryId, t) => {
+  // if this entry is to a group, construct one
+  let groupPromise = Promise.resolve(null)
+  if (entry.group) {
+    groupPromise = Group.create({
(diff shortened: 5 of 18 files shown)


backend/db/migrations/20171116200117-create-group.js
```diff
@@ -22,5 +22,5 @@ export function up (queryInterface, Sequelize) {
 }
 
 export function down (queryInterface) {
-  return queryInterface.dropTable('group')
+  return queryInterface.dropTable('groups')
```
robmcl4: This was a minor mistake in a previous migration, might as well fix it

backend/models/image.js
```diff
@@ -21,6 +21,14 @@ export default sequelize.define('image', {
...
     allowNull: false,
     notEmpty: true
+  },
+  createdAt: {
+    allowNull: false,
+    type: DataTypes.DATE
+  },
+  updatedAt: {
+    allowNull: false,
+    type: DataTypes.DATE
```
robmcl4: we forgot these, might as well add

backend/models/other.js
```diff
@@ -7,6 +7,14 @@ export default sequelize.define('other', {
...
     type: DataTypes.STRING,
     allowNull: true
+  },
+  createdAt: {
+    allowNull: false,
+    type: DataTypes.DATE
+  },
+  updatedAt: {
+    allowNull: false,
+    type: DataTypes.DATE
```
robmcl4: forgot these, might as well add

backend/models/video.js
```diff
@@ -13,6 +13,14 @@ export default sequelize.define('video', {
...
     allowNull: false,
     notEmpty: true
+  },
+  createdAt: {
+    allowNull: false,
+    type: DataTypes.DATE
+  },
+  updatedAt: {
+    allowNull: false,
+    type: DataTypes.DATE
```
robmcl4: forgot these, might as well add

backend/resolvers/queries/userQuery.js
```diff
@@ -3,7 +3,7 @@ import { UserError } from 'graphql-errors'
 import { STUDENT, ADMIN, JUDGE } from '../../constants'
 
 export function user (_, args, req) {
-  const isRequestingOwnUser = req.auth.id !== undefined && req.auth.id === args.id
+  const isRequestingOwnUser = req.auth.username !== undefined && req.auth.username === args.id
```
robmcl4: seemed to be broken.... fixed a bit but the schema still expects `id` in the args, which we should think about

backend/test/resolvers/judge.js
```diff
@@ -7,6 +7,11 @@ import { fakeUser } from '../factories'
 
 describe('Judge Resolvers', function () {
   describe('Judge Creation Resolver', function () {
+    it('Does now allow non-admins to add judge', function () {
```
robmcl4: shamelessly added to boost coverage above 80%, along with a number of others

backend/test/resolvers/userQuery.js
```diff
@@ -10,7 +10,7 @@ describe('User Queries', function () {
   describe('User query', function () {
     it('Allows users to find themeselves', function (done) {
       fakeUser({username: 'user1'}).then((u) => {
-        user('', {id: u.username}, {auth: {type: 'STUDENT', id: 'user1'}})
+        user('', {id: u.username}, {auth: {type: 'STUDENT', username: 'user1'}})
```
robmcl4: partially fixed for new auth expected, but again this `id` thing seems odd

backend/test/resolvers/entry.js
```diff
@@ -1,21 +1,58 @@
 import { expect } from 'chai'
 
 import Entry from '../../models/entry'
+import Group from '../../models/group'
 import { createPhoto, createVideo, createOtherMedia } from '../../resolvers/mutations/entry'
 import { fakeUser, fakeShow } from '../factories'
 
+const standardEntry = (user, show) => ({
```
kayladavis: Should we create a factory for Entry?

backend/resolvers/queries/userQuery.js
kayladavis: Graphql IDs don't necessarily mean int ideas, just unique ids. The ID type is serialized in the same way as a String. So, I don't think I'd want to stop identifying the user's pk as a graphql id. I rather change the pk for user to a normal id.

backend/resolvers/mutations/entry.js
```diff
@@ -5,9 +5,35 @@ import Entry from '../../models/entry'
...
+  if (entry.group) {
+    groupPromise = Group.create({
+      name: entry.group.name,
+      creatorUsername: entry.group.creatorUsername,
+      participants: entry.group.participants
+    }, {transaction: t})
+  }
+  return groupPromise
+    .then(group => {
+      // remove the `group` attribute and replace it with the `groupId`
```
kayladavis: Please add a more fleshed out description of what is happening here and why

backend/test/resolvers/entry.js
robmcl4: probably not necessary here, since the point of testing this method is to create the Entry object itself

PR #130, opened 2018-01-27
[WIP] Remove rebass on front page, centering not working

chtinahow:
image
Currently looking like this. I'm having issues getting it to center in the middle of the page correctly

### DIFF
--- a/frontend/src/Home/Page.js
+++ b/frontend/src/Home/Page.js
@@ -1,5 +1,5 @@
 import React from 'react'
-import { Row, Col } from 'reactstrap'
+import { Container, Row, Col } from 'reactstrap'
 import { Flex } from 'rebass'
 import styled from 'styled-components'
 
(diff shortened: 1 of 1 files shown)


chtinahow:
New view: 
image

billdybas:
I just glanced at your code - didn't checkout your branch.

Yea, this is on the right track. We want to remove all instances where we use `rebass` to minimize dependencies and keep styling more cohesive.

This means, anywhere we use `rebass` features like `<Flex>` and `<Box>` (from `grid-styled`), convert them into the equivalent Reactstrap (i.e. `<Row>`, `<Col>`) or raw CSS code.

When you were talking about this in person, you made it seem like you were hung up on choosing between Reactstrap vs. raw CSS (i.e. flexbox). Reactstrap's (aka Bootstrap 4's) grid system _**is**_ flexbox – the `<Row>` and `<Col>` components are just flexbox wrappers that provide some convenience and consistency around styling at different breakpoints. 

`<Row>` and `<Col>` should be the starting point when creating a new layout, but feel free to deviate as necessary and drop down to raw CSS if the layout doesn't fit perfectly into the Bootstrap mold. The point of using Bootstrap in this project is to make layout generally consistent across the codebase and to reduce having to write all the flexbox layout code ourselves.

I honestly don't care how you go about switching over from `rebass` provided the pages look and respond the same as before (and you're _not_ using CSS grids, per our previous discussion – still too experimental). Use your best judgement and just ship, ship, ship!

http://getbootstrap.com/docs/4.0/layout/grid/
https://reactstrap.github.io/components/layout/

frontend/src/Home/Page.js
```diff
@@ -1,5 +1,5 @@
 import React from 'react'
-import { Row, Col } from 'reactstrap'
+import { Container, Row, Col } from 'reactstrap'
 import { Flex } from 'rebass'
```
robmcl4: unused import?

frontend/src/Home/Page.js
```diff
@@ -37,16 +43,17 @@ const LoginButton = styled.a`
     text-decoration: none;
   }
 `
-
+// .col .col-sm-12 .col-md-6 .col-md-offset-3
```
kayladavis: Why are these comments here?

billdybas:
Closes #157

Issue #129, opened 2018-01-23
pdf upload / serving


robmcl4:

chtinahow:
Add a path for PDF

billdybas:
Move both images and PDFs into a folder `uploads`

uploads/
-> images/
-> pdfs/

PR #132, opened 2018-01-30
File upload post auth and pdf upload 

kayladavis:
This pr covers authing file posts (making sure the user is a student or admin) and also pdf upload.


kayladavis:
also connects to #129, but zenhub doesn't seem to allow attaching more than one issue

backend/server.js
```diff
@@ -11,7 +11,9 @@ import models from './models'
 import router from './routes'
 import schema from './schema'
 import parseJwtUser from './middleware/parseJwtUser'
-import imageUploader from './upload'
+import { imageUploader, pdfUploader } from './upload'
+import { allowedToSubmit } from './helpers/submission'
```
robmcl4: seems like an unused import

backend/test/routes/fileUpload.js
```diff
@@ -0,0 +1,126 @@
+import request from 'supertest'
```
robmcl4: Please also re-format to 2 spaces

backend/upload.js
```diff
@@ -4,46 +4,75 @@ import uuidv1 from 'uuid/v1'
...
-export default function uploader (req, res) {
-  upload(req, res, (err) => {
-    if (err) {
-      if(err.message === 'File too large'){
-        return res.status(413).json({error: 'File too large'})
-      }
-      return res.status(500).json({error: 'Failed to Upload'})
-    }
+function pdfTypeFilter (req, file, cb, mimetype) {
+  if (file.mimetype !== 'application/pdf') { // verify image is a jpg
```
robmcl4: jpg -> pdf

backend/upload.js
```diff
@@ -4,46 +4,75 @@ import uuidv1 from 'uuid/v1'
 import nconf from './config'
```
robmcl4: I think this file should be moved to `backend/routes/upload.js`, which allows us to use the Router module in combination with a small edit of `backend/routes/index.js`. As a bonus, we can leave `backend/server.js` fairly pristine.

Router is pretty clean to use:

```javascript
const router = Router()

router.route('/path/to/thing')
  .get(
    middlewareFunc, // probably your uploadAuth function goes here
    (req, res, next) => imageUpload( /* ... */ )
  )

export default router
```

backend/server.js
```diff
@@ -34,7 +36,19 @@ if (config.get('NODE_ENV') !== 'production') {
   app.get('/graphiql', graphiqlExpress({ endpointURL: '/graphql' }))
 }
 
-app.use('/static', express.static(path.join(__dirname, 'images')))
-router.post('/static/upload', imageUploader) // TODO: Require Authentication
+function uploadAuth(req, res, next) {
```
robmcl4: this middleware function feels odd defined here, maybe move it to `backend/middleware` or, if you're changing the `upload.js` file to use Router (see other comment), into there?

PR #133, opened 2018-01-30
Sequelize migration/model fix to match schema expectations for EntryInput

kayladavis:


backend/db/migrations/20171116292217-create-entry.js
```diff
@@ -43,17 +43,20 @@ export function up (queryInterface, Sequelize) {
     },
     title: {
       type: Sequelize.STRING,
+      defaultValue: 'untitled',
```
billdybas: Can you change this to a capital 'u' – "Untitled"

PR #140, opened 2018-01-31
Adds bulk zip downloading for a single show

robmcl4:
Adds bulk zip download for a show.

This is currently unauthenticated, which will be addressed in the future.

Route is `/zips/{showId}` and prompts the user-agent to download `{show name}.zip` file.

The zip file contains structure:
```
invited/Last First - Title.jpg
invited/Last First - Title (1).jpg
invited/GroupName - Title.jpg
not invited/Last First - Title.jpg
```


backend/routes/download.js
```diff
@@ -0,0 +1,259 @@
...
+import config from '../config'
+import sequelize from '../config/sequelize';
+
+const readFileAsync = promisify(fs.readFile)
+
+const IMAGE_DIR = config.get('upload:imageDir')
+
+const router = Router()
+
+router.route('/zips/:showId')
```
kayladavis: Just a general comment on this function. This looks like the right approach to the problem, but I kind of want it to be split into a few smaller functions. I think it's fine and we don't need to change it now though.

PR #142, opened 2018-02-01
Student Submission Forms

billdybas:
Adds Student Submission Forms to the Student app.

Still To Do:
- [x] Add a 'Success!' Modal after the submission is made and is successful
- [x] ~~~If there is an issue w/ the form, display an error and don't clear the form (this is primarily for the Video Submission – letting the backend validate the URL is from YouTube or Vimeo)~~~ => See #148
- [x] Making the submission for a particular show (and updating routes and page title (eg. "You are submitting to Honors Show 2018"))
- [x] Adding in the Group workflow
- [x] Determine if we need `moreCopies` on Video and Other Media submissions
- [x] Make `mediaType` on Photo Submission a [react-select](https://github.com/JedWatson/react-select) `Creatable`
- [x] ~~~Better style the Dropzone previews~~~ => See #160
- [x] UX on 'Back' buttons + 'Cancel' submission
- [x] ~~~Convert the other forms using `redux-form` to `formik` (Probably a separate PR)~~~ => See #156
- [x] Fix wording of forms based on Nanette's feedback (Artist Comment => optional, Physical Art")
- [x] If there's an uploaded image and someone goes to another page, clear the preview

~~~The Dropzone's will not work until #132 is merged in, as it uses the new upload paths defined in that PR.~~~ Works!

Additionally, we should determine if we need to have some way to delete uploads, as you can upload something to the Dropzone, click on it and upload again, causing the first upload to not ever be associated w/ a submission and forever take up disk space. => See #159


frontend/src/Student/components/OtherMediaSubmissionForm.js
```diff
@@ -0,0 +1,287 @@
...
+                setFieldValue(name, this.props.previewFile.path)
+              })
+              break
+            case 'image/jpeg':
+              handleImageUpload(file).then(() => {
+                // Need to use 'this.props' here to get the most up-to-date value – 'previewFile' above will be out-of-date
+                setFieldValue(name, this.props.previewFile.path)
+              })
+              break
+            default:
```
chtinahow: what are we doing about the video case?

billdybas: Students who want to submit a YouTube or Vimeo video will choose the "Video" Submission Form – this is the "Other Media" form where students can submit an image or PDF of something like a book or physical object.

We don't need to worry about videos on this form, as students cannot upload video files.

frontend/src/Student/components/OtherMediaSubmissionForm.js
```diff
@@ -0,0 +1,400 @@
...
+              console.error(`Unknown File Type: ${file.type}`)
```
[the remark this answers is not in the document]
billdybas: So, this code should _technically_ never be reached. We only `accept='application/pdf,image/jpeg'`, so there would need to be issues w/ the Dropzone or someone would have to be modifying our internals through dev tools.

frontend/src/Student/pages/Submit.js
```diff
@@ -1,18 +1,26 @@
...
+      <Route exact path='/submit/photo' render={() => <PhotoSubmissionForm forShow={props.forShow} />} />
+      <Route exact path='/submit/video' render={() => <VideoSubmissionForm forShow={props.forShow} />} />
+      <Route exact path='/submit/other' render={() => <OtherMediaSubmissionForm forShow={props.forShow} />} />
+    </Switch>
   </Container>
 )
 
-export default Submit
+const mapStateToProps = (state) => ({
+  forShow: { id: 1, name: 'Honors Show 2017' } // TODO: Get from Redux
```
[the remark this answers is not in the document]
billdybas: Yes, this is intentional – not really a big issue. When @chtinahow has the Student dashboard complete, when you click "Make Submission" on a particular show, we'll either pass that show's id and name as props or access them through Redux. Since she'll have already queried for the show to display it on the dashboard, we'll just need to pass that data along.

If you want to make an Issue for integrating our two features together, feel free.

frontend/src/Student/components/SuccessModal.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react'
+import { Modal, ModalBody } from 'reactstrap'
+import FaCheckCircle from 'babel-loader!react-icons/fa/check-circle'
+
+const SuccessModal = (props) => (
+  <Modal
+    isOpen={props.isOpen}
```
chtinahow: I guess this wold be for cleanliness sake, but why not break the styles out into their own divs?

billdybas: iirc, this was a bit hacky just to get the styles to work like I wanted. I couldn't use reacstrap's `<ModalHeader />` component because the HTML it generates made it impossible to center the check icon.

If by "break[ing] the styles out into their own divs" you mean make them styled-components, I don't see any benefit over how the styles are right now – there's no component re-use. If we need more modals like this in the future, I could see making this a generic component and passing in props like the background color, icon, and body copy.

kayladavis:
I think overall this is solid. The others covered any concern I have. So lgtm.

frontend/src/Student/components/OtherMediaSubmissionForm.js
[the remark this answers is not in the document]
billdybas: I remember I also added this case because in `react-dropzone`'s docs, they mention that MIME type determination isn't reliable across platforms – if and end user is having issues uploading a file, someone would be able to diagnose the issue by having the end user open their console and see what the error message is.

See the `accept` prop: https://react-dropzone.js.org/

Issue #136, opened 2018-01-30
Summary information on judging


robmcl4:
On the backend, we want:

* Entries have a `score` calculated property
*  modify `votes({judgeUsername: ..., showId: ...}): [Vote]` to make `judgeUsername` optional if you're an admin
* Score is averaged over the total votes

PR #151, opened 2018-02-06
Adding vote mutation, migration, and query

kayladavis:


backend/resolvers/mutations/vote.js
```diff
@@ -0,0 +1,58 @@
...
+export function vote(_, args, req) {
+  // Make sure judge is voting as themself
+  const isRequestingOwnJudgeUser = req.auth.username !== undefined &&
+    req.auth.type === JUDGE && req.auth.username === args.input.judgeUsername
+  if (!isRequestingOwnJudgeUser) {
+    throw new UserError('Permission Denied')
+  }
+  // Make sure judge is allowed to vote on this entry
+  const input = args.input
+  return judgeIsAllowedToVote(input).spread((allowedToVote, errMsg) => {
```
robmcl4: using above changes, this can be simplified:

```javascript
return judgeIsAllowedToVote(input)
  .then(() => {
    // no error handling required, since the promise rejects before this block
  })
```

backend/resolvers/mutations/vote.js
```diff
@@ -0,0 +1,58 @@
...
+  return User.findById(input.judgeUsername).then(judge => {
+    return Entry.findById(input.entryId).then(entry => {
+      if (!entry) {
+        return [false, 'Cannot find entry']
+      }
+      return judge.getShows().then(assignedShows => {
+        // Make sure the array contains the show this entry is part of
+        const allowedToVote = assignedShows.filter(show => show.id === entry.showId)
+        if (allowedToVote.length === 0) {
+          return [false, 'Judge is not allowed to vote on this entry']
```
robmcl4: Use

```javascript
return Promise.reject(new UserError('Judge is not allowed to vote on this entry'))
```

or simply

```javascript
throw new UserError('Judge is not allowed to vote on this entry')
```

backend/resolvers/queries/voteQuery.js
```diff
@@ -0,0 +1,25 @@
+import Vote from '../../models/vote'
+import User from '../../models/user'
+import Entry from '../../models/entry'
+
+import { UserError } from 'graphql-errors'
+import { ADMIN, JUDGE } from '../../constants'
+
+export function votes (_, args, req) {
+  const isRequestingOwnJudgeUser = req.auth.username !== undefined &&
+  req.auth.type === JUDGE && req.auth.username === args.judgeUsername
```
robmcl4: nit: indent

backend/schema.js
```diff
@@ -74,6 +74,12 @@ type Vote {
     entry: Entry
```
robmcl4: these properties are not resolving currently

To add:

1. add a `backend/resolvers/types/voteType.js` which defines `judge` and `entry` properties somehow, see [`backend/resolvers/types/groupType.js`](https://github.com/robmcl4/A-B-S-T-R-A-C-T-I-O-N/blob/master/backend/resolvers/types/groupType.js) as an example. You may also find it useful to define some helper methods on `Vote`.
2. add the above file to `backend/resolvers/index.js`

we haven't exactly figured out to test this behavior yet, so don't worry it's easy to overlook :/

kayladavis:
referencing #136  because commit 26b94e0 completes it.

PR #153, opened 2018-02-08
add download token

robmcl4:
Adds a download token, also refactors token signing and parsing into a helper module. Includes tests of functionality.

### DIFF
--- a/backend/routes/download.js
+++ b/backend/routes/download.js
@@ -10,9 +10,10 @@ import Group from '../models/group'
 import Image from '../models/image'
 import Show from '../models/show'
 import User from '../models/user'
-import { IMAGE_ENTRY } from '../constants'
+import { IMAGE_ENTRY, ADMIN } from '../constants'
 import config from '../config'
-import sequelize from '../config/sequelize';
+import sequelize from '../config/sequelize'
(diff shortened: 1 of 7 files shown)


backend/routes/download.js
```diff
@@ -10,20 +10,30 @@ import Group from '../models/group'
...
 const router = Router()
 
-router.route('/zips/:showId')
-  .get((req, res, next) => {
-    // TODO figure out how to do auth
+const ensureAdminDownloadToken = (req, res, next) => {
+  const token = req.query.token
+  parseToken(token, (err, decoded) => {
+    if (err || decoded.type !== ADMIN) {
+      res.status(401).send('{"message": "permission denied"}')
```
kayladavis: I think the identifier here should be consistent with /routes/auth. In auth you use `error: message`. Your non-success status messages are created different ways in this file.

robmcl4: good point, will do

robmcl4: I realized that it makes no sense to render json here anyway, since this text will be displayed on a user's browser if they are not permitted to access this... I changed it to HTML in 88b6abf

Issue #156, opened 2018-02-10
Tech Debt: Finish Converting Forms to Formik


billdybas:
Convert the rest of the frontend forms from `redux-form` to `formik` and remove `redux-form` from our `package.json` dependencies.

robmcl4:
To Bill: datepicker the entry period in creating a show does not allow me to select ranges, closes after you select start date. Firefox v58.0.1

PR #161, opened 2018-02-11
Remove Rebass from the project

### LINKED ISSUE #155: Create Show Form Throws Errors w/ Datepicker
For some reason, there are errors thrown which prevent the page from rendering. We currently have the component commented out.

1. Find why the errors are being thrown
2. Fix it so that we can render the form

robmcl4:
Removes Rebass.

Also restructures how the home page scales, hiding the accent image on `xs` screens and fixing the ever-present horizontal scrollbar.

Also tweaks how buttons look on the Show card (sorry, couldn't help myself)

Album of changes: https://imgur.com/a/2vd1w 

Closes #155
Closes #130 (Duplicate PR)

### DIFF
--- a/frontend/src/Home/Page.js
+++ b/frontend/src/Home/Page.js
@@ -4,13 +4,11 @@ import styled from 'styled-components'
 
 import splash from 'assets/splash-page.jpg'
 
-const SplashImage = styled.img`
+const SplashImage = styled.div`
   background-image: url(${props => props.src});
-  object-fit: cover;
+  background-size: cover;
   height: 100vh;
-  width: 41%;
-  position: fixed;
-  right: 0;
+  width: 100%;
 `
 
 const PageTitle = styled.h1`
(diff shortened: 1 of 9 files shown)


frontend/src/Home/Page.js
```diff
@@ -23,6 +23,13 @@ const ButtonContainer = styled.div`
   width: 250px;
 `
 
+const LoginContainer = styled.div`
+  justify-content: center;
+  min-height: 75vh;
+  display: flex;
+  flex-direction: column;
```
chtinahow: This will definitively give you the vertical layout you need for the login area, but it's not horizontally centered correctly. You'll probably want to add a separate div or something else around the login container to center it.

chtinahow: I might just be pedantic about it, so don't work super hard on it if you can't get it centered

frontend/src/Home/Page.js
```diff
@@ -1,14 +1,14 @@
...
-import { Flex } from 'rebass'
+import { Container, Row, Col } from 'reactstrap'
 import styled from 'styled-components'
 
 import splash from 'assets/splash-page.jpg'
 
 const SplashImage = styled.img`
   background-image: url(${props => props.src});
-  background-size: cover;
+  object-fit: cover;
```
billdybas: Not supported in IE 11

frontend/src/index.js
```diff
@@ -1,3 +1,10 @@
+// Initialize Global Imports
```
billdybas: Why did you move these?

robmcl4: fixes things

frontend/src/Home/Page.js
robmcl4: fixed centering in future commit

billdybas:
LGTM. 🚀 

Only qualm (and technically a non-requirement) I have is on `CreateShowForm` I would move the date pickers underneath the other form fields on smaller screens – right now they collapse horizontally. I can do this when I get to #156.

PR #163, opened 2018-02-13
Toggle between admin and judge role

kayladavis:
In this PR I added actions/reducers to toggle judge/admin roles and produce the admin or judge view based on the active performing role.

I also realized a bug with the way I wrote voting (admins need to vote too). So that's fixed an tested.

### DIFF
--- a/frontend/src/Admin/components/NavBar.js
+++ b/frontend/src/Admin/components/NavBar.js
@@ -4,7 +4,6 @@ import { Link } from 'react-router-dom'
 import { connect } from 'react-redux'
 import { Collapse, Navbar, NavbarToggler, Nav, NavItem, NavLink } from 'reactstrap'
 import styled from 'styled-components'
-import JudgePage from '../../Judge/Page'
 import { switchToJudge } from '../../shared/actions'
 
 const NavBarContainer = styled.div`
@@ -40,7 +39,7 @@ class NavBar extends Component {
           <Collapse isOpen={this.state.isOpen} navbar>
             <Nav className="ml-auto" navbar>
               <NavItem>
-                <Link to='/' className='nav-link' onClick={this.props.switchToJudge}>Switch to: Judge</Link>
+                <NavLink tag={Link} to='/' onClick={this.props.switchToJudge}>Switch to: Judge</NavLink>
               </NavItem>
               <NavItem>
                 <NavLink tag={Link} to='/judges'>Judges</NavLink>
--- a/frontend/src/Admin/components/NavBar.js
+++ b/frontend/src/Admin/components/NavBar.js
@@ -1,23 +1,27 @@
 import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 import { Link } from 'react-router-dom'
+import { connect } from 'react-redux'
 import { Collapse, Navbar, NavbarToggler, Nav, NavItem, NavLink } from 'reactstrap'
 import styled from 'styled-components'
+import JudgePage from '../../Judge/Page'
+import { switchToJudge } from '../../shared/actions'
 
 const NavBarContainer = styled.div`
   margin-bottom: 25px;
 `
 
-export default class NavBar extends Component {
+class NavBar extends Component {
   static propTypes = {
-    logout: PropTypes.func.isRequired
+    logout: PropTypes.func.isRequired,
+    switchToJudge: PropTypes.func.isRequired,
   }
 
   constructor (props) {
     super(props)
 
     this.state = {
-      isOpen: false
+      isOpen: false,
     }
   }
 
@@ -36,6 +40,9 @@ export default class NavBar extends Component {
           <Collapse isOpen={this.state.isOpen} navbar>
             <Nav className="ml-auto" navbar>
               <NavItem>
+                <Link to='/' className='nav-link' onClick={this.props.switchToJudge}>Switch to: Judge</Link>
+              </NavItem>
+              <NavItem>
                 <NavLink tag={Link} to='/judges'>Judges</NavLink>
               </NavItem>
               <NavItem>
@@ -51,3 +58,13 @@ export default class NavBar extends Component {
     )
   }
 }
+
+
+const mapDispatchToProps = (dispatch) => ({
+  switchToJudge: () => dispatch(switchToJudge())
+})
+
+export default connect(
+  () => ({}),
+  mapDispatchToProps
+)(NavBar) 
--- a/frontend/src/Admin/components/NavBar.js
+++ b/frontend/src/Admin/components/NavBar.js
@@ -4,13 +4,12 @@ import { Link } from 'react-router-dom'
 import { connect } from 'react-redux'
 import { Collapse, Navbar, NavbarToggler, Nav, NavItem, NavLink } from 'reactstrap'
 import styled from 'styled-components'
-import { switchToJudge } from '../../shared/actions'
 
 const NavBarContainer = styled.div`
   margin-bottom: 25px;
 `
 
-class NavBar extends Component {
+export default class NavBar extends Component {
   static propTypes = {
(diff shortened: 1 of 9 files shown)


frontend/src/Admin/components/NavBar.js
```diff
@@ -35,6 +38,9 @@ export default class NavBar extends Component {
           <NavbarToggler onClick={this.toggle} />
           <Collapse isOpen={this.state.isOpen} navbar>
             <Nav className="ml-auto" navbar>
+              <NavItem>
+                <NavLink tag={Link} to='/' onClick={this.props.switchToJudge}>Switch to: Judge</NavLink>
```
chtinahow: Quick thing,I think that the Switch to: Judge looks weird compared to the rest of the links on the navbar. No big deal if others disagree

frontend/src/Admin/components/NavBar.js
```diff
@@ -51,3 +57,13 @@ export default class NavBar extends Component {
     )
   }
 }
+
+
+const mapDispatchToProps = (dispatch) => ({
+  switchToJudge: () => dispatch(switchToJudge())
```
robmcl4: Can you stick this instead in [`frontend/src/Admin/containers/NavBar.js`](https://github.com/robmcl4/A-B-S-T-R-A-C-T-I-O-N/blob/master/frontend/src/Admin/containers/NavBar.js#L8), we've been leaving the "components" folder free of redux hookup

frontend/src/Judge/components/NavBar.js
```diff
@@ -45,3 +51,12 @@ export default class NavBar extends Component {
     )
   }
 }
+
+const mapDispatchToProps = (dispatch) => ({
+  switchToAdmin: () => dispatch(switchToAdmin())
```
robmcl4: same as before, can you stick the redux hookup in the "containers" folder

frontend/src/shared/reducers.js
```diff
@@ -33,6 +33,21 @@ const auth = (state = {}, action) => {
...
+      return newState  
+    default:
+      return state
+  }
+}
+
 export default combineReducers({
-  auth
+  auth,
+  performingRole
```
robmcl4: this smells a bit odd to me... I feel like `performingRole` should be a a key in the `auth` reducer that maps to a string, such that auth looks like:

```json
{
   "token": "...",
   "user": {
     "username": "souperManne",
     "firstName": "Clark",
     "lastName": "Kent",
     "type": "ADMIN",
     "createdAt": "...",
     "iat": 1516390788
  },
  "performingRole": "JUDGE"
}
```

... which I think might be a bit cleaner for everything, especially considering the state currently doesn't clear when you logout (bug or feature tho?)

frontend/src/shared/components/LoginSwitch.js
```diff
@@ -56,12 +63,14 @@ class LoginSwitch extends Component {
...
-  user: state.shared.auth.user
+  user: state.shared.auth.user,
+  performingRole: state.shared.performingRole
 })
 
 const mapDispatchToProps = (dispatch) => ({
   shouldLogin,
-  login: () => dispatch(login())
+  login: () => dispatch(login()),
+
```
robmcl4: nit: extra line

frontend/src/shared/reducers.js
```diff
@@ -33,6 +33,21 @@ const auth = (state = {}, action) => {
   }
 }
 
+const performingRole = (state = {type: 'ADMIN'}, action) => {
+  let newState = state
+  switch (action.type) {
+    case actions.JUDGE_ROLE:
+      newState.type = 'JUDGE'
+      return newState
```
robmcl4:
```javascript
return {type: 'JUDGE'}
```

frontend/src/shared/reducers.js
kayladavis: I think moving it to auth could work. This task was mostly me learning to use redux and I just wasn't thinking of designing it that way when I started.

frontend/src/Admin/components/NavBar.js
```diff
@@ -1,6 +1,7 @@
 import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 import { Link } from 'react-router-dom'
+import { connect } from 'react-redux'
```
billdybas: You can remove this, now.

frontend/src/Admin/components/NavBar.js
```diff
@@ -10,14 +11,15 @@ const NavBarContainer = styled.div`
...
+    logout: PropTypes.func.isRequired,
+    switchToJudge: PropTypes.func.isRequired,
   }
 
   constructor (props) {
     super(props)
 
     this.state = {
-      isOpen: false
+      isOpen: false,
```
billdybas: nit: extra comma

frontend/src/Judge/components/NavBar.js
```diff
@@ -1,6 +1,7 @@
 import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 import { Link } from 'react-router-dom'
+import { connect } from 'react-redux'
```
billdybas: You can remove this, now.

frontend/src/shared/components/LoginSwitch.js
```diff
@@ -33,14 +35,20 @@ class LoginSwitch extends Component {
     const {
       user
     } = this.props
+    const performingRole = this.props.performingRole
```
billdybas: You can destructure this like the line above it:

```js
const {
  user,
  performingRole
} = this.props
```

frontend/src/shared/components/LoginSwitch.js
```diff
@@ -56,12 +64,13 @@ class LoginSwitch extends Component {
...
 const mapStateToProps = (state) => ({
-  user: state.shared.auth.user
+  user: state.shared.auth.user,
+  performingRole: state.shared.auth.performingRole
 })
 
 const mapDispatchToProps = (dispatch) => ({
   shouldLogin,
-  login: () => dispatch(login())
+  login: () => dispatch(login()),
```
billdybas: nit: extra comma

frontend/src/shared/actions.js
```diff
@@ -3,6 +3,8 @@ import { push } from 'connected-react-router'
 
 export const LOGIN_USER = 'LOGIN_USER'
 export const LOGOUT_USER = 'LOGOUT_USER'
+export const JUDGE_ROLE = 'JUDGE_ROLE'
+export const ADMIN_ROLE = 'ADMIN_ROLE'
```
billdybas: I think I would prefer these actions to be named:

```js
export const SWITCH_TO_JUDGE = 'SWITCH_TO_JUDGE'
export const SWITCH_TO_ADMIN = 'SWITCH_TO_ADMIN'
```

Don't forget to update the references in other files.

frontend/src/Admin/components/NavBar.js
billdybas: I think I would rename it to "View as Judge"

frontend/src/Judge/components/NavBar.js
```diff
@@ -35,6 +37,9 @@ export default class NavBar extends Component {
           <NavbarToggler onClick={this.toggle} />
           <Collapse isOpen={this.state.isOpen} navbar>
             <Nav className="ml-auto" navbar>
+              <NavItem>
+                <NavLink tag={Link} to='/' onClick={this.props.switchToAdmin}>Switch to: Admin</NavLink>
```
billdybas: I would rename this to "View as Admin"

billdybas: Also, I would add some sort of visual separation between this and the other links, say a `|` symbol in between them.

Issue #164, opened 2018-02-13
modify entries query to allow students to find their own entries


robmcl4:
modify `entries(showId: ID): [Entry]` to take a second argument `studentUsername: string` which allows students to find all of their own entries. When this second argument is supplied, showId is optional.

robmcl4:
wishlist ideal implementation:

query to return `[Show]` of all the shows a student has submitted to, each show's `entries` embedded property is populated by only the entries which a student has submitted

PR #166, opened 2018-02-15
Allowing students to get their own entries by searching entries or searching shows

kayladavis:
To handle the issues described in #164 I modified two queries: shows and entries.

Both queries now have an optional studentUsername param. The entries query solves the first comment's problem. And the shows query solves the wishlist comment's problem.

### DIFF
--- a/backend/models/user.js
+++ b/backend/models/user.js
@@ -4,7 +4,7 @@ import sequelize from '../config/sequelize'
 import { STUDENT, ADMIN, JUDGE } from '../constants'
 import Entry from './entry';
 
-const User =  sequelize.define('user', {
+const User = sequelize.define('user', {
   username: {
     type: DataTypes.STRING,
     allowNull: false,
@@ -32,36 +32,25 @@ const User =  sequelize.define('user', {
 })
 
 User.prototype.getGroups = function getGroups() {
-  if(this.type !== STUDENT) {
+  if (this.type !== STUDENT) {
     return Promise.resolve([])
   }
   // Find all groups where the creator is this user
-  return Group.findAll({where: {creatorUsername: this.username}})
+  return Group.findAll({ where: { creatorUsername: this.username } })
 }
 
-User.prototype.getOwnAndGroupEntries = function getOwnAndGroupEntries() {
-  if(this.type !== STUDENT) {
+User.prototype.getOwnAndGroupEntries = function getOwnAndGroupEntries(showIds = null) {
+  if (this.type !== STUDENT) {
     return Promise.resolve([])
   }
   // Find all entries where this user is the submitter (group and self)
   return this.getGroups()
     .then((groups) => {
       const groupIds = groups.map(group => group.id)
-      if (groupIds.length > 0) {
-        return Entry.findAll({
-          where: {
-            $or: [
-              {
-                groupId: groupIds
-              }, 
-              {
-                studentUsername: this.username
-              }
-            ]
-          }
-        })
+      if (showIds) {
+        return Entry.findAll({ where: { $or: [{ groupId: groupIds }, { studentUsername: this.username }], showId: showIds } })
       } else {
-        return Entry.findAll({ where: { studentUsername: this.username } })
+        return Entry.findAll({ where: { $or: [{ groupId: groupIds }, { studentUsername: this.username }] } })
       }
     })
 }
--- a/backend/models/user.js
+++ b/backend/models/user.js
@@ -2,6 +2,7 @@ import Group from './group'
 import DataTypes from 'sequelize'
 import sequelize from '../config/sequelize'
 import { STUDENT, ADMIN, JUDGE } from '../constants'
+import Entry from './entry';
 
 const User =  sequelize.define('user', {
   username: {
@@ -36,6 +37,33 @@ User.prototype.getGroups = function getGroups() {
   }
   // Find all groups where the creator is this user
   return Group.findAll({where: {creatorUsername: this.username}})
+}
+
+User.prototype.getOwnAndGroupEntries = function getOwnAndGroupEntries() {
+  if(this.type !== STUDENT) {
+    return Promise.resolve([])
+  }
+  // Find all entries where this user is the submitter (group and self)
+  return this.getGroups()
+    .then((groups) => {
+      const groupIds = groups.map(group => group.id)
+      if (groupIds.length > 0) {
+        return Entry.findAll({
+          where: {
+            $or: [
+              {
+                groupId: groupIds
+              }, 
+              {
+                studentUsername: this.username
+              }
+            ]
+          }
+        })
+      } else {
+        return Entry.findAll({ where: { studentUsername: this.username } })
+      }
+    })
 }
 
 export default User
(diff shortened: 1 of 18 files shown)


robmcl4:
As per our conversation earlier, here's my thoughts on what _might_ be helpful here (drafted in github, some minor typos may follow):

1. Refactor the `Entry` polymorphic properties so we can stop using the `.dataValues` hack (+1 consistency)

In the Entry type resolver we do the following to find and cache the specific entry object for later field extraction:

```javascript
__resolveType (data, context, info) {
  if (data.entryType === IMAGE_ENTRY) {
    context.image = data.getImage() // <------------- new
    return info.schema.getType('Photo')
  } else if (data.entryType === VIDEO_ENTRY) {
    context.video = data.getVideo() // <------------- new
    return info.schema.getType('Video')
  } else if (data.entryType === OTHER_ENTRY) {
    context.other = data.getOther() // <------------- new
    return info.schema.getType('OtherMedia')
  }
```

We'll need three new `resolvers/types/{photo|video|other}Type.js` files so we can tell GraphQL what the values are for the subtype-specific fields. For example, `photoType.js` stub:

```javascript
export default {
  Photo: {
    path (obj, args, context) {
      return context.image.then(photo => photo.path)
    },
    horizDimInch (obj, args, context) {
      return context.image.then(photo => photo.horizDimInch)
    }
  },
  // ...
}
```

And that should let us remove every instance of `Object.assign(entry, image.dataValues)` which would be _amazing_, plus bonus points for this just auto-magically filling in those properties even when Entry is an embedded property (unlike right now)

2. We can also tell a Show's type resolver to only retrieve the entries that are relevant.

Surprisingly, there are more arguments to the type resolver. We can do the following:

```javascript
export default {
  Show: {
    entries (show, args, req) {
      if (req.auth.type === ADMIN || req.auth.type === JUDGE) {
         // return all the entries in this show
      }
      else if (req.auth.type === STUDENT) {
        // return only the student's relevant entries ...... this is probably where you can put
        // the monster sequelize query
      }
    }
  }
}
```

and beyond that.... I think we might not actually need the modified `entries(...)` query, if the above meets the frontend's needs

So sorry for not getting this clarified sooner, I feel bad for having you spin your wheels too much!

backend/models/user.js
```diff
@@ -28,3 +30,40 @@ export default sequelize.define('user', {
...
+  // Find all groups where the creator is this user
+  return Group.findAll({where: {creatorUsername: this.username}})
+}
+
+User.prototype.getOwnAndGroupEntries = function getOwnAndGroupEntries() {
+  if(this.type !== STUDENT) {
+    return Promise.resolve([])
+  }
+  // Find all entries where this user is the submitter (group and self)
+  return this.getGroups()
```
robmcl4: ugly, but useful: thanks for this

backend/resolvers/queries/showQuery.js
```diff
@@ -1,15 +1,50 @@
...
+    // Get all the shows the student has been on 
+    // (including as group creator), through entries
+    return User.findById(args.studentUsername).then((student) => {
+      return student.getOwnAndGroupEntries()
+        // Find all connected shows
+        .then((entries) => {
+          const showIds = entries.map(entry => entry.showId)
+          // If order doesn't matter just return all the shows
+          if (!args.orderBy) {
+            return Show.findAll({ where: { id: showIds } })
```
robmcl4: just wondering, does this automatically do a `WHERE thing IN (a, b, c)` query? I've seen sequelize that uses the $in operator

backend/resolvers/types/entryType.js
```diff
@@ -1,19 +1,33 @@
...
-    },
-    score(entry) {
-      return entry.getScore()
-    },
     __resolveType (data, context, info) {
+      // Identifies for GraphQL which concrete instance of Entry this object is.
+      // Here we also prefetch the associated photo/video/other object so that
+      // their specific type resolvers can pull out those properties
       if (data.entryType === IMAGE_ENTRY) {
+        context.image = data.getImage()
```
robmcl4: can prolly remove these

backend/resolvers/types/entryType.js
```diff
@@ -1,19 +1,33 @@
...
-    group (entry) {
-      return entry.getGroup()
-    },
-    score(entry) {
-      return entry.getScore()
-    },
     __resolveType (data, context, info) {
+      // Identifies for GraphQL which concrete instance of Entry this object is.
+      // Here we also prefetch the associated photo/video/other object so that
+      // their specific type resolvers can pull out those properties
```
robmcl4: hmmmmmm can remove the last 2 lines of this comment since we should also remove a bit of what it describes below

backend/resolvers/types/showType.js
```diff
@@ -2,6 +2,9 @@ export default {
   Show: {
     judges (show) {
       return show.getUsers()
+    },
+    entries (show) {
+      return show.getEntries()
```
robmcl4: Here is where we can get interesting..... I believe we can do something like the following, which prevents us from leaking other peoples' entries to the front-end.

```javascript
entries (show, _, req) {
  if (req.auth.type === ADMIN || req.auth.type === JUDGE)
    return show.getEntries()
  else
    // return only this user's relevant entries
    return User.findById(req.auth.username)
      .then(u => u.getOwnAndGroupEntries())
      .then(entries => entries.filter(e => e.showId === show.id)
}

// NOTE THAT THIS IS JUST AN EXAMPLE:
// I bet there's a neat sequelize way to do this without
// throwing out unneeded shows, but whatever this
// is probably fast enough
```

backend/test/resolvers/showQuery.js
```diff
@@ -18,26 +19,68 @@ describe('Show Queries', function () {
...
+                    }
+                  }
+                  `, 
+                  {type: 'STUDENT', username: u.username}
+                ).then(result => {
+                  expect(result.data.shows.length).to.equal(2)
+                  expect(result.data.shows[0].entries.length).to.equal(1)
+                  done()
+                })
+                // shows('', { studentUsername: u.username },
```
robmcl4: whoops... remove commented-out stuff

backend/resolvers/queries/showQuery.js
billdybas: [Looks like it should](https://github.com/sequelize/sequelize/blob/091abb48ea1aa459fb6955536ae499d47845e342/lib/model.js#L1492)

kayladavis: yup. see here: http://docs.sequelizejs.com/manual/tutorial/models-usage.html#-findall-search-for-multiple-elements-in-the-database

backend/resolvers/types/showType.js
```diff
@@ -1,7 +1,21 @@
...
       return show.getUsers()
+    },
+    entries(show, _, req) {
+      // Admins and judges should see all entries on a show
+      if (req.auth.type === ADMIN || req.auth.type === JUDGE) {
+        return show.getEntries()
+      } else {
+        return User.findById(req.auth.username)
+          .then((user) => {
+            return user.getOwnAndGroupEntries(show.id)
```
robmcl4: great solution!

Issue #169, opened 2018-02-15
As a judge, I want "refreshable" URLs


billdybas:
The URLs in the judging interface should be structured so that if you refresh the page on a show or submission, you see the same show or submission – you aren't taken to the start of judging or the app entrypoint.

billdybas:
Because judges might want to share URLs with each other, I think the URL should somehow reference a particular show and submission (perhaps by id – no information leak here since the judge can see all submissions, though we might have to think about excluded submissions).

The order of the shuffled submissions + ids would be stored in Redux, and going "left/back" or "right/forward" would just get the element at the next index in that list and view it. But the id in the URL need not be sequential.

For example, say for show (id = 1), if we have submissions, `[1, 2, 3, 4, 5]`, and they get shuffled for a judge to `[2, 4, 3, 1, 5]`, then the judge could see:

`/show/1?submission=2`
`/show/1?submission=2`

Another URL format could be `/show/<show_id>/submission/<submission_id>`.

Would maybe also need to add permissions so that Judges can't see submissions from shows they haven't been assigned to.

billdybas:
We could also add a `public_id` to entries and shows which is something like a [hashid](http://hashids.org/) - like a 7 char unique string.

Then anywhere on the frontend we need to reference ids, we use the `public_id`.

This prevents id enumeration issues - like a judge guessing the id of an entry excluded from judging (though we should still design some experience here – probably a not found page). Additionally, if in  the future the system were to be upgraded to use Postgres or another db, discrepancies w/ internally stored ids wouldn't affect previously shared / bookmarked URLs.

PR #187, opened 2018-02-16
Tina/new student show

### LINKED ISSUE #182: As a Student I would like to see a new show that I have not submitted to
Acceptance Criteria:
Show Name 
Description(Nice to have)
Submission Deadline
Entry Cap
New Submission Button(links to the form)
All the backend Hookup will be here
  This includes the judging status

https://imgur.com/a/mRMe0

chtinahow:
Resolves #182 
This does the graphql query for all of the student side for the status.
This includes the new prototype of what Rob made shown here: 
image


frontend/src/Student/components/ShowCard.js
```diff
@@ -0,0 +1,55 @@
...
+       <Link to={`/submit/${props.id}`} className='d-block w-100 h-100'>
+         <FaPlusCircle size='3em' />
+         <h5>New Submission</h5>
+       </Link>
+      </Col>
+    </Row>
+  </Card>
+)
+
+ShowCard.propTypes = {
```
billdybas: Instead of having all these props passed in shallowly, I would encapsulate them in a `show` prop:

```js
ShowCard.propTypes = {
  show: PropTypes.shape({
    id: PropTypes.string,
    ...
    judgingEnd: PropTypes.string
  }).isRequired
}
```

See comment about `entries` below

frontend/src/Student/components/ShowCard.js
```diff
@@ -0,0 +1,55 @@
+import React from 'react'
+import styled from 'styled-components'
+import { Link } from 'react-router-dom'
+import PropTypes from 'prop-types'
+import Moment from 'react-moment'
+import FaPlusCircle from 'babel-loader!react-icons/fa/plus-circle'
+import { Button, ButtonGroup, Row, Col } from 'reactstrap'
+
+const Card = styled.div`
```
billdybas: Visually, this feels too narrow. `min-height: 250px` seems to work well. You'll have to recenter the "New Submission" button.

frontend/src/Student/components/ShowCard.js
```diff
@@ -0,0 +1,55 @@
...
+
+const ShowCard = (props) => (
+  <Card>
+    <Row>
+      <Col>
+        <h2>{props.name}</h2>
+      </Col>
+      <Col className="text-right">
+      <div>
+        <h5>{props.entries ? props.entries.length : 0}/{props.entryCap} submissions</h5>
```
billdybas: With the `defaultProps` below, `props.show.entries.length` will be zero when there are no submissions, so you could simplify this to just `{props.show.entries.length}/{props.show.entryCap}`

Also, see comment below about "Submissions" vs. "Entries"

frontend/src/Student/components/ShowCard.js
```diff
@@ -0,0 +1,55 @@
...
+  margin-top: 15px;
+  padding: 10px;
+  width: 100%;
+`
+
+const ShowCard = (props) => (
+  <Card>
+    <Row>
+      <Col>
+        <h2>{props.name}</h2>
```
billdybas: I would align this text so the baseline is the same as "Submission Ends"

frontend/src/Student/components/OtherMediaSubmissionForm.js
```diff
@@ -30,10 +30,6 @@ class OtherSubmissionForm extends Component {
     user: PropTypes.shape({
       username: PropTypes.string
     }).isRequired,
-    forShow: PropTypes.shape({
```
billdybas: If you're removing this in each of these files, you need to add in the `propTypes` for `data`

frontend/src/Student/queries/showName.graphql
```diff
@@ -0,0 +1,6 @@
+query Show($id: ID!) {
```
billdybas: It feels weird to me making a query just for the name, but I guess it works: if we're coming to this page from the chooser, apollo probably has the data cached. If we're loading this page new, we'll need to fetch the name.

Still getting used to GraphQL ¯\\\_(ツ)\_/¯

frontend/src/Student/components/Shows.js
```diff
@@ -0,0 +1,19 @@
...
+const Shows = ({shows, loading}) => (
+  <div>
+    {loading ? null : shows.map((show) => (
+      <ShowCard key={show.id} {...show} />
+    ))}
+  </div>
+)
+
+Shows.propTypes = {
+  shows: PropTypes.array,
```
billdybas: I'd make this required and add a `defaultProp` of `[]`

frontend/src/Student/pages/Submit.js
```diff
@@ -11,16 +11,14 @@ import OtherMediaSubmissionForm from '../containers/OtherMediaSubmissionForm'
 const Submit = (props) => (
   <Container>
     <Switch>
-      <Route exact path='/submit' component={SubmissionFormChooser} />
-      <Route exact path='/submit/photo' render={() => <PhotoSubmissionForm forShow={props.forShow} />} />
-      <Route exact path='/submit/video' render={() => <VideoSubmissionForm forShow={props.forShow} />} />
-      <Route exact path='/submit/other' render={() => <OtherMediaSubmissionForm forShow={props.forShow} />} />
+      <Route exact path='/submit/:id' component={SubmissionFormChooser} />
```
billdybas: Not a _huge_ fan of this URL format here.

Thoughts about doing:

`/submit?to=<public_id>`
`/submit/photo?to=<public_id>`
`/submit/video?to=<public_id>`
`/submit/other?to=<public_id>`

or 

`/submit?to=<public_id>`
`/submit?to=<public_id>&type=photo`
`/submit?to=<public_id>&type=video`
`/submit?to=<public_id>&type=other`

Without the `to` param or an invalid/non-authorized `public_id`, you'd be redirected to your dashboard.

See #169 for my thoughts on using a `public_id`. We could just use the show's id for now.

frontend/src/Student/components/ShowCard.js
```diff
@@ -0,0 +1,55 @@
...
+      </Col>
+      <Col className="text-right">
+      <div>
+        <h5>{props.entries ? props.entries.length : 0}/{props.entryCap} submissions</h5>
+      </div>
+      <div>
+        Submission Ends: <Moment format='YYYY/MM/DD'>{props.entryEnd}</Moment>
```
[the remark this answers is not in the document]
chtinahow: I think they should be submissions as that's what Nanette usually uses, but @robmcl4 @kayladavis opinions?

frontend/src/Student/components/ShowCard.js
chtinahow: Oh true, good catch

frontend/src/Student/components/ShowCard.js
chtinahow: You want it centered with  Accepting Submissions Until correct?

frontend/src/Student/components/ShowCard.js
chtinahow: So doing so without the ternery causes a null if the entries are completely empty shown here in this error: 
image

frontend/src/Student/components/ShowCard.js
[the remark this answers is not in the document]
chtinahow: Also, I don't think react-moment would be valid anymore if we decided to go with the glance able format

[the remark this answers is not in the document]
kayladavis: I agree with "submissions". At some point we just chose entries as types because it was slightly shorter of a word.

Also, I agree with Bill for the date format and wording.

frontend/src/Student/containers/PhotoSubmissionForm.js
```diff
@@ -6,10 +6,11 @@ import { uploadImage, clearPreview } from '../actions'
 
 import PhotoSubmissionForm from '../components/PhotoSubmissionForm'
 import CreatePhotoEntry from '../mutations/createPhotoEntry.graphql'
+import ShowName from '../queries/showName.graphql'
 
-const mapStateToProps = (state) => ({
+const mapStateToProps = (state, ownProps) => ({
   previewImage: state.student.ui.submission.previewFile || {},
-  user: state.shared.auth.user
+  user: state.shared.auth.user,
```
kayladavis: Nit: extra comma

frontend/src/Student/queries/shows.graphql
```diff
@@ -0,0 +1,15 @@
...
+  shows(orderBy: { sort: "entryStart", direction: DESC }) {
+    id
+    name
+    entryStart
+    entryEnd
+    judgingStart
+    judgingEnd
+    entryCap
+    entries {
+      id
```
kayladavis: Will we need more than just the id?

chtinahow: I think so? Would the invited status be associated with the entries?

frontend/src/Student/components/Shows.js
chtinahow: If there are no so associated with the student, would they be able to still log in?

frontend/src/Student/pages/Submit.js
chtinahow: Probably going with /submit?to=<public_id>
/submit/photo?to=<public_id>
/submit/video?to=<public_id>
/submit/other?to=<public_id>

frontend/src/Student/pages/Submit.js
```diff
@@ -1,38 +1,25 @@
...
-import { connect } from 'react-redux'
 
 import SubmissionFormChooser from '../components/SubmissionFormChooser'
 import PhotoSubmissionForm from '../containers/PhotoSubmissionForm'
 import VideoSubmissionForm from '../containers/VideoSubmissionForm'
 import OtherMediaSubmissionForm from '../containers/OtherMediaSubmissionForm'
 
-const Submit = props => (
+const addIdPropFromQueryParams = () => {
+  return window.location.search.split('=')[1]
```
billdybas: This is not the most robust way to do this, as any additional query parameters will cause issues.

eg. `/submit?foo=bar&to=1` causes the buttons to link to `/submit/<type>?to=bar`

Additionally, no check is done to make sure a show w/ the id exists, so you can view the chooser for any id and when you attempt to view the submission forms for that non-existent show, the app crashes.

Ideally, we would parse the query string specifically for the param `to` and  check if the id in the `to` query param is a valid show which this student can submit to, otherwise, render a Not Found page.

For MVP purposes, this'll work, but I'm making a story for having more robust query string parsing.

PR #188, opened 2018-02-17
refactor Entry instance fields into type resolvers, plus full-query tests

robmcl4:


chtinahow:
@robmcl4  You have a CI failure just by the way

robmcl4:
@chtinahow sorry, yes, this is WIP -- goes onto Kayla's PR; still needs some fixup

PR #189, opened 2018-02-20
README

billdybas:
Adds a top level README about our project, team, architecture.

- [x] Only section I have left is the "Process" section.

Let me know if there's something I missed we should add or if anything should be re-worded / changed / removed.

When we add in more tools (eg. image processing) we can add them then. And when we deploy, we can write those instructions.

- [x] Probably should also add a general "Here's how to maintain Node" section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,1 +1,89 @@
 # Gallery Gateway
+
+![Gallery Gateway Logo](https://user-images.githubusercontent.com/13719429/36440348-8d156e20-163d-11e8-8462-dfe4ba840a41.png)
+
+## About
+
+### Motivation
+
+The [RIT](https://www.rit.edu/) [School of Photographic Arts and Sciences (SPAS)](http://cias.rit.edu/schools/photographic-arts-sciences) holds an annual art exhibition highlighting the top works of RIT student artists. Students can submit works for consideration during an open call, and a panel of RIT faculty jurors vote on which art will be displayed in the gallery at this show. Currently, SPAS has been paying for and using a software as a service product, [CaFÉ](https://www.callforentry.org/), to facilitate this process; however, frustrations with CaFÉ’s difficult-to-use interface and poor user-experience for both administrative and student workflows has led them to seek custom replacement software which better fits their use case and work process.
+
+Gallery Gateway is this custom replacement software.
+
+### Team
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -0,0 +1,89 @@
...
+- [GraphQL](http://graphql.org/)
+- [Apollo GraphQL](https://www.apollographql.com/)
+- [Apollo Developers Blog](https://dev-blog.apollodata.com/)
+- [The GitHub GraphQL API](https://githubengineering.com/the-github-graphql-api/)
+- [From REST to GraphQL](https://0x2a.sh/from-rest-to-graphql-b4e95e94c26b) by Jacob Gillespie
+- [awesome-graphql](https://github.com/chentsulin/awesome-graphql)
+
+## Development
+
+You'll need to be running both the frontend and backend for development. Check out each of their README's for further instructions.
```
chtinahow: I would probably reiterate the instructions here or link to the front end read me and then backend readme

kayladavis: I'd lean towards linking.

README.md
```diff
@@ -0,0 +1,89 @@
...
+Both the frontend and backend are written in JavaScript.
+
+Frontend:
+
+- [React](https://reactjs.org/) for view rendering
+- [Redux](https://redux.js.org/) for state management
+- [React Router](https://reacttraining.com/react-router/) v4 for page routing
+- [Apollo Client](https://github.com/apollographql/apollo-client) for GraphQL querying and data caching
+- [Boostrap](http://getbootstrap.com/) v4 (via [reactstrap](https://reactstrap.github.io/)) & [styled-components](https://www.styled-components.com/) for styling
+- [Formik](https://github.com/jaredpalmer/formik) w/ [yup](https://github.com/jquense/yup) for forms
```
robmcl4: mention webpack somehow... seems a biggie

billdybas: Oh, yes, definitely. Slipped my mind completely.

README.md
billdybas: I'm going to say neither. You're in the top level of the repo, it's not hard to click on the  `backend` or `frontend` folders and read their README. Additionally, links inside our own repo will require maintenance – eg. when we move the repo to the org, we'll have to update the links.

PR #191, opened 2018-02-22
Fixing backend linting errors

kayladavis:
Running `yarn run format` on the frontend doesn't produce any changes


backend/package.json
```diff
@@ -15,6 +15,14 @@
     "sourceMap": false,
     "instrument": false
   },
+  "eslintConfig": {
+    "env": {
+      "browser": true,
```
robmcl4: erm.... I have no idea how ESlint works but this looks odd to me

backend/package.json
```diff
@@ -74,4 +83,4 @@
     "standard": "^10.0.3",
     "supertest": "^3.0.0"
   }
-}
+}
```
robmcl4: nit: every other file ends in a newline (UNIX style) ... we should also end this file in a newline

backend/package.json
billdybas: [Specifying Environments](https://eslint.org/docs/user-guide/configuring#specifying-environments)

"An environment defines global variables that are predefined."

`"broswer": true,` should be removed, as browser globals are not available in Node

the other config options are fine, though.

billdybas:
@kayladavis Try `npm run format` on the frontend. I get at least 40 files changed.

backend/package.json
kayladavis: oops, meant to remove that. I was pulling in stuff from the config page and forgot.

PR #193, opened 2018-02-22
[WIP] show student's entries on their dashboard

robmcl4:
IN PROGRESS DO NOT MERGE

builds on top of Tina's PR #187

### DIFF
--- a/frontend/src/Student/components/ShowCard.js
+++ b/frontend/src/Student/components/ShowCard.js
@@ -1,0 +1,55 @@
+import React from 'react'
+import styled from 'styled-components'
+import { Link } from 'react-router-dom'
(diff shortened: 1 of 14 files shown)


frontend/src/Student/components/ShowCard.js
```diff
@@ -0,0 +1,132 @@
...
+`
+
+const EntryNoThumbContainer = styled.div`
+  height: auto;
+  width: auto;
+  max-height: 10em;
+  max-width: 100%;
+`
+
+const renderEntry = (props, entry) => {
```
chtinahow: Should we possibly break these out into their own components? Showcard may get kind of bulky otherwise....

Issue #173, opened 2018-02-15
As a judge, I would like to know if I haven't been assigned to any shows


billdybas:
Make a 0 state on the Judge Dashboard.

chtinahow:
"Not currently assigned to a show"

PR #194, opened 2018-02-25
Adding assigned show cards for judges

kayladavis:
If a judge is assigned  to shows they are listed in the dashboard. Currently all shows are listed because that's the easiest way to retrieve the data.

I also added the "No assigned show" message (solves #173) when there is no show the user has been assigned to.

### DIFF
--- a/backend/resolvers/queries/userQuery.js
+++ b/backend/resolvers/queries/userQuery.js
@@ -10,6 +10,13 @@ export function user (_, args, req) {
   return User.findById(args.id)
 }
 
+export function self(_, __,req) {
+  if (!req.auth.username){
+    throw new UserError('Permission Denied')
+  }
+  return User.findById(req.auth.username)
+}
+
 export function users (_, args, req) {
   if (req.auth.type !== ADMIN) {
     throw new UserError('Permission Denied')
--- a/frontend/src/Judge/components/ShowCard.js
+++ b/frontend/src/Judge/components/ShowCard.js
@@ -16,7 +16,6 @@ const ButtonContainer = styled.div`
   position: absolute;
   right: 0;
   bottom: 0;
-  width: '150px';
 `
 
 const ShowCard = props => (
@@ -35,6 +34,7 @@ const ShowCard = props => (
     <Col>
       <ButtonContainer>
         <Button
+          size='lg'
           style={{ cursor: 'pointer' }}
           tag={Link}
           to={`show/${props.id}/vote`}
--- a/frontend/src/Judge/components/ShowCard.js
+++ b/frontend/src/Judge/components/ShowCard.js
@@ -3,7 +3,7 @@ import styled from 'styled-components'
 import { Link } from 'react-router-dom'
 import PropTypes from 'prop-types'
 import Moment from 'react-moment'
-import { Button, ButtonGroup, Row, Col } from 'reactstrap'
+import { Button, Row, Col } from 'reactstrap'
 
 const Card = styled.div`
   background-color: #f8f9fa;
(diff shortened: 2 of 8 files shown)


frontend/src/Judge/components/Shows.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+import ShowCard from '../components/ShowCard'
+
+const Shows = ({ user, loading }) => (
+  <div>
+    {loading ? null :
+    user.shows.length > 0 ? user.shows.map(show => <ShowCard key={show.id} {...show} />) :
+    "You are not currently assinged to a show"
```
chtinahow: I would center this instead of having it off to the side to stay with the flow when cards do show up

frontend/src/Judge/components/ShowCard.js
```diff
@@ -0,0 +1,56 @@
...
+`
+
+const ShowCard = props => (
+  <Card>
+    <h2>
+      <Link to={`show/${props.id}`}>{props.name}</Link>
+    </h2>
+    <Col>
+      <dl>
+        <dt>Judging Ends:</dt>
```
chtinahow: This is the additional code you were talking about right?

frontend/src/Judge/components/ShowCard.js
```diff
@@ -0,0 +1,56 @@
...
+  position: absolute;
+  right: 0;
+  bottom: 0;
+  width: '150px'
+`
+
+const ShowCard = props => (
+  <Card>
+    <h2>
+      <Link to={`show/${props.id}`}>{props.name}</Link>
```
chtinahow: This was changed as per request of Bill's to be a query param, I believe the link to would look like this:  <Link to={`/submit?to=${forShow.id}`}>

frontend/src/Judge/components/ShowCard.js
```diff
@@ -0,0 +1,56 @@
...
+          <Moment format='YYYY/MM/DD'>{props.judgingEnd}</Moment>
+        </dd>
+      </dl>
+    </Col>
+    <Col>
+      <ButtonContainer>
+        <Button
+          style={{ cursor: 'pointer' }}
+          tag={Link}
+          to={`judge/vote`} // TODO: Figure out where they go to vote
```
chtinahow: This might also be a query param that goes to  <Link to={`/vote?to=${forShow.id}`}>
And also putting in their username or something for the unique shuffle? This is probably a later thing.

frontend/src/Judge/components/Shows.js
billdybas: I'd also center the text (horizontally and vertically). Also, `user.shows.length ? ...` should be fine (0 is falsey)

frontend/src/Judge/components/ShowCard.js
billdybas: I'm thinking the URL formats we'll use on the Judge app are:

`/show/<show_id>/` => shows the review page for a show (the list view of all entries and what the judge voted)

`/show/<show_id>/vote/` =>  **(use this here)** Prepares the voting queue for the particular show. Fetches all entries, creates a shuffle order, and then renders the first unvoted entry in the voting interface (updating the URL to the format below). If the voting queue has already been prepared, accessing this route will recalculate the last unvoted entry and go to it.

`/show/<show_id>/vote?entry=<entry_id>` => shows the voting interface w/ the particular entry

frontend/src/Judge/components/ShowCard.js
kayladavis: The entire show card? Yeah.

frontend/src/Judge/components/ShowCard.js
kayladavis: But this is for judges. They won't be submitting to a show. I'll just remove the link I guess.

frontend/src/Judge/components/ShowCard.js
chtinahow: I agree with bill on this one

frontend/src/Judge/components/ShowCard.js
kayladavis: Never mind. Bill's comment `/show/<show_id>/` should still work here

frontend/src/Judge/components/Shows.js
```diff
@@ -0,0 +1,31 @@
...
+import styled from 'styled-components'
+import PropTypes from 'prop-types'
+
+import ShowCard from '../components/ShowCard'
+
+const NoShowsContainer = styled.div`
+  position: fixed;
+  top: 50%;
+  left: 50%;
+  transform: translate(-50%, -50%);
```
billdybas: Haven't looked at this visually, but a pro-tip, with flexbox, centering things is _real_ easy:

```css
display: flex;
align-items: center;
justify-content: center;
```

[Reference](https://philipwalton.github.io/solved-by-flexbox/demos/vertical-centering/)

kayladavis: That only horizontally aligns it: 
image
vs the vertical allignment you asked for: 
image

backend/resolvers/queries/userQuery.js
```diff
@@ -10,6 +10,13 @@ export function user (_, args, req) {
   return User.findById(args.id)
 }
 
+export function self(_, __,req) {
```
billdybas: nit: the second param is a double `_` but could be a single `_`

frontend/src/Judge/components/ShowCard.js
```diff
@@ -0,0 +1,56 @@
...
+import { Button, Row, Col } from 'reactstrap'
+
+const Card = styled.div`
+  background-color: #f8f9fa;
+  border-radius: 5px;
+  margin-top: 15px;
+  padding: 10px;
+  width: 100%;
+`
+const ButtonContainer = styled.div`
```
chtinahow: I would make the start button a little bit bigger since there's so much space on the card itself

PR #198, opened 2018-02-28
Judges see only active assigned show cards

kayladavis:
This PR makes it so only active (between entry start and judging end) shows are displayed.
Here's an example of the change (I have two shows running that are assigned, one is in entry phase, the other judging): 
image

updated (c450af9): 
image

Ignore any of the changes in frontend/student. They were altered when I ran the format script. I don't know why it's not running on push as the pkg json says it should.

### DIFF
--- a/backend/resolvers/types/userType.js
+++ b/backend/resolvers/types/userType.js
@@ -1,7 +1,8 @@
 export default {
   User: {
+    // Judges (and Admins) have shows
     shows (user) {
-      return user.getShows()
+      return user.getShows({order: [['judgingStart', 'DESC']]})
     }
   }
 }
--- a/backend/resolvers/types/userType.js
+++ b/backend/resolvers/types/userType.js
@@ -1,12 +1,14 @@
 export default {
   User: {
-    // Judges (and Admins) have shows
+    // Judges (and Admins) are assigned to shows, Students are not
     shows (user, args) {
-      // Check if given date is in a show range
+      if (user.type === 'STUDENT') {
+        return []
+      }
+      // Check if given date is before any show ends
       if (args.date) {
         return user.getShows({
           where: {
-            entryStart: {lte: args.date},
             judgingEnd: {gte: args.date}
           },
           order: [['judgingStart', 'DESC']]
--- a/backend/resolvers/types/userType.js
+++ b/backend/resolvers/types/userType.js
@@ -1,7 +1,18 @@
 export default {
   User: {
     // Judges (and Admins) have shows
-    shows (user) {
+    shows (user, args) {
+      // Check if given date is in a show range
+      if (args.date) {
+        return user.getShows({
+          where: {
+            entryStart: {lte: args.date},
+            judgingEnd: {gte: args.date}
+          },
+          order: [['judgingStart', 'DESC']]
+        })
+      }
+      // For no args just find all shows this user is assigned to
       return user.getShows({order: [['judgingStart', 'DESC']]})
     }
   }
--- a/frontend/src/Judge/components/ShowCard.js
+++ b/frontend/src/Judge/components/ShowCard.js
@@ -19,63 +19,51 @@ const ButtonContainer = styled.div`
   bottom: 0;
 `
 
-function renderCardContents (props) {
-  const now = moment()
-  const judgingStart = moment(props.judgingStart)
-  if (now.isBefore(judgingStart)) {
-    return renderBeforeJudging(props)
-  } else {
-    return renderDuringJudging(props)
-  }
-}
-function renderDuringJudging (props) {
-  return (
-    <div>
-      <Col>
-        <dl>
-          <dt>Judging Progress:</dt>
-          <dd>
-            {props.ownVotes.length} / {props.entries.length}
-          </dd>
-          <dt>Judging Ends:</dt>
-          <dd>
-            <Moment format='YYYY/MM/DD'>{props.judgingEnd}</Moment>
-          </dd>
-        </dl>
-      </Col>
-      <Col>
-        <ButtonContainer>
-          <Button
-            size='lg'
-            style={{ cursor: 'pointer' }}
-            tag={Link}
-            to={`show/${props.id}/vote`}
-            // TODO: Conditionally change the text
-          >
-            {props.ownVotes.length === 0 && props.entries.length === 0
-              ? 'Start'
-              : props.ownVotes.length === props.entries.length
-                ? 'Review'
-                : 'Resume'}
-          </Button>
-        </ButtonContainer>
-      </Col>
-    </div>
-  )
-}
+const DuringJudging = ({ ownVotes, entries, judgingEnd, id }) => (
+  <div>
+    <Col>
+      <dl>
+        <dt>Judging Progress:</dt>
+        <dd>
+          {ownVotes.length} / {entries.length}
+        </dd>
+        <dt>Judging Ends:</dt>
+        <dd>
+          <Moment format='YYYY/MM/DD'>{judgingEnd}</Moment>
+        </dd>
+      </dl>
+    </Col>
+    <Col>
+      <ButtonContainer>
+        <Button
+          size='lg'
+          style={{ cursor: 'pointer' }}
+          tag={Link}
+          to={`show/${id}/vote`}
+        >
+          {ownVotes.length === 0
+            ? 'Start'
+            : ownVotes.length === entries.length
+              ? 'Review'
+              : 'Resume'}
+        </Button>
+      </ButtonContainer>
+    </Col>
+  </div>
+)
 
-function renderBeforeJudging (props) {
-  return (
+const BeforeJudging = ({ judgingStart, judgingEnd }) => (
+  <div>
     <Col>
       <div> Judging has not started yet. Come back to vote soon! </div>
       <dl>
         <dt>Judging Starts:</dt>
         <dd>
-          <Moment format='YYYY/MM/DD'>{props.judgingStart}</Moment>
+          <Moment format='YYYY/MM/DD'>{judgingStart}</Moment>
         </dd>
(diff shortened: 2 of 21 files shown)


backend/resolvers/types/userType.js
```diff
@@ -1,7 +1,18 @@
 export default {
   User: {
     // Judges (and Admins) have shows
-    shows (user) {
+    shows (user, args) {
```
robmcl4: err.... I think this is put in the wrong place? Probably should be in `backend/resolvers/queries/showQuery.js`?

frontend/src/Judge/components/ShowCard.js
```diff
@@ -18,39 +19,93 @@ const ButtonContainer = styled.div`
...
-    </h2>
+function renderCardContents (props) {
+  const now = moment()
+  const judgingStart = moment(props.judgingStart)
+  if (now.isBefore(judgingStart)) {
+    return renderBeforeJudging(props)
+  } else {
+    return renderDuringJudging(props)
+  }
+}
```
robmcl4: nit: add newline between functions

frontend/src/Judge/components/ShowCard.js
```diff
@@ -18,39 +19,93 @@ const ButtonContainer = styled.div`
   bottom: 0;
 `
 
-const ShowCard = props => (
-  <Card>
-    <h2>
-      <Link to={`show/${props.id}`}>{props.name}</Link>
-    </h2>
+function renderCardContents (props) {
```
robmcl4:
```javascript
const CardContents = props => {
```

frontend/src/Judge/components/ShowCard.js
```diff
@@ -18,39 +19,93 @@ const ButtonContainer = styled.div`
...
+  const now = moment()
+  const judgingStart = moment(props.judgingStart)
+  if (now.isBefore(judgingStart)) {
+    return renderBeforeJudging(props)
```
robmcl4:
```javascript
return <BeforeJudging {...props} />
```

same w/ `DuringJudging` found below

frontend/src/Judge/components/ShowCard.js
```diff
@@ -18,39 +19,93 @@ const ButtonContainer = styled.div`
...
+        </dl>
+      </Col>
+      <Col>
+        <ButtonContainer>
+          <Button
+            size='lg'
+            style={{ cursor: 'pointer' }}
+            tag={Link}
+            to={`show/${props.id}/vote`}
+            // TODO: Conditionally change the text
```
robmcl4: this Todo looks done?

frontend/src/Judge/components/ShowCard.js
```diff
@@ -18,39 +19,93 @@ const ButtonContainer = styled.div`
...
+function renderCardContents (props) {
+  const now = moment()
+  const judgingStart = moment(props.judgingStart)
+  if (now.isBefore(judgingStart)) {
+    return renderBeforeJudging(props)
+  } else {
+    return renderDuringJudging(props)
+  }
+}
+function renderDuringJudging (props) {
```
robmcl4:
```javascript
const DuringJudging = ({
    ownVotes,
    entries,
    judgingEnd
  }) => (
  <div> ...
)
```

frontend/src/Judge/components/ShowCard.js
```diff
@@ -18,39 +19,93 @@ const ButtonContainer = styled.div`
...
+      </Col>
+    </div>
+  )
+}
+
+function renderBeforeJudging (props) {
+  return (
     <Col>
+      <div> Judging has not started yet. Come back to vote soon! </div>
       <dl>
```
robmcl4: I feel like in the future we can be more stylish than a `<dl>` but i'm okay with this for now unless there are large objections

frontend/src/Judge/components/ShowCard.js
```diff
@@ -18,39 +19,93 @@ const ButtonContainer = styled.div`
...
+                ? 'Review'
+                : 'Resume'}
+          </Button>
+        </ButtonContainer>
+      </Col>
+    </div>
+  )
+}
+
+function renderBeforeJudging (props) {
```
robmcl4:
```javascript
const BeforeJudging = ({
    ownVotes,
    entries,
    judgingStart,
    judgingEnd
  }) => (
  <div> ...
)
```

frontend/src/Judge/components/ShowCard.js
```diff
@@ -18,39 +19,93 @@ const ButtonContainer = styled.div`
...
-    </Col>
+  )
+}
+
+const ShowCard = props => (
+  <Card>
+    <h2>
+      <Link to={`show/${props.id}`}>{props.name}</Link>
+    </h2>
+    {renderCardContents(props)}
```
robmcl4: `<CardContents {...props} />`

backend/resolvers/types/userType.js
```diff
@@ -1,7 +1,18 @@
 export default {
   User: {
     // Judges (and Admins) have shows
```
billdybas: I'm confused about the `(and Admins)` part of this comment and how it relates to the rest of this function.

If any user, regardless of type, makes this query, they'll get the shows ordered by `'judgingStart` unless they provide a date, no? Remind me how Students and Admins are associated to shows – would they need a different default? 

Additionally, for Judges, it seems useful to know if you've been assigned to any future show, not just those that are in the submission or judging phase, so that you can plan to come back at a future date.

frontend/src/Judge/components/ShowCard.js
billdybas: Style-wise I think most cards will need a visual overhaul (eg. Admin dashboard), so for MVP-purposes, it's alright. Once we have the MVP and get feedback, we should have finalized the data that should go on all the cards, so it'll be easier to architect the information to make them ＡＥＳＴＨＥＴＩＣ

frontend/src/Judge/components/Shows.js
```diff
@@ -18,7 +18,7 @@ const Shows = ({ user, loading }) => (
       user.shows.map(show => <ShowCard key={show.id} {...show} />)
     ) : (
       <NoShowsContainer>
-        You are not currently assinged to a show
+        You are not currently assinged to an active show
```
billdybas: See comment above: "any future shows"

frontend/src/Judge/components/ShowCard.js
```diff
@@ -18,39 +19,93 @@ const ButtonContainer = styled.div`
...
+      <Col>
+        <ButtonContainer>
+          <Button
+            size='lg'
+            style={{ cursor: 'pointer' }}
+            tag={Link}
+            to={`show/${props.id}/vote`}
+            // TODO: Conditionally change the text
+          >
+            {props.ownVotes.length === 0 && props.entries.length === 0
```
billdybas: Why the `&& props.entries.length === 0`?

frontend/src/Judge/components/ShowCard.js
billdybas: I would just get rid of this function. See my comment on your comment below.

frontend/src/Judge/components/ShowCard.js
billdybas: Since we're changing the above functions to stateless components, I would just change this line to be:

```jsx
{moment().isBefore(moment(props.judgingStart))
  ? <BeforeJudging {...props} />
  : <DuringJudging {...props} />}

File: frontend/src/Judge/components/ShowCard.js [hunk 13]
```diff
@@ -18,39 +19,93 @@ const ButtonContainer = styled.div`
...
   </Card>
 )
 
 ShowCard.propTypes = {
   id: PropTypes.string.isRequired,
   name: PropTypes.string.isRequired,
-  judgingEnd: PropTypes.string.isRequired
+  judgingStart: PropTypes.string.isRequired,
+  judgingEnd: PropTypes.string.isRequired,
+  entries: PropTypes.arrayOf(
```
billdybas: Provide `defaultProps` for this and `ownVotes`:

```js
ShowCard.defaultProps = {
  entries: [],
  ownVotes: []
}
```

backend/resolvers/types/userType.js
kayladavis: No, because this hooks onto the user_show table, which is only used in assigning judges. There's no student show association formally. We could add a check for students and show them something else. Students and admins should be able to to vote because Nanette still wants to function as a judge. 

As far as what they see I specifically asked about the range. I can open it to any future show if you believe that's better

billdybas: Ah, ok. 

Technically, though, if a relation was made between a Student and a Show (like manually editing the db), calling this query would return that row, right? I'm just trying to understand how architecturally things fit together – it just feels weird that in the schema there are functions which are only relevant to one type of user but not another – I'm wondering if there's a better schema design (i.e. for that type of relation – probably not worth the time / effort to change things / re-engineer now, but I'm just thinking about future GraphQL projects).

For Students, might make sense to do an empty object or "Permission Denied" since they're not allowed to vote.

I'm sorry if I mis-understood what you were asking about at our meeting about the date range. When I was testing this out yesterday, I had to make new shows in the future, and I was a bit confused that even after I assigned judges to the shows, I couldn't view them as a judge. UX-wise I think showing all future shows the judges have been assigned to is a better workflow – say Nanette emails them that they're supposed to vote on an upcoming show, but that show hasn't entered the submission phase yet, might confuse judges who log on and says they haven't been assigned to any shows.

kayladavis: You're right. I'll open the date range. I can also make students always get an empty array.

kayladavis: Slightly related. we probably should beef up our seed to make it a little easier to test.

Issue #184, opened 2018-02-15
As a Student, I would like to see the judging results of my entries


chtinahow:
Acceptance Criteria:
Status bar Underneath of photos(Green Accepted, Black/Grey Not accepted)
No Make Submission Button
Entries Shown
Show Name
Submission date as "Closed"

kayladavis:
Acceptance Criteria:
Status bar Underneath of photos(Green Accepted, Black/Grey Not accepted)
No Make Submission Button
~~Entries Shown~~ solved by rob
~~Show Name~~ solved by rob
Submission date as "Closed"

PR #201, opened 2018-03-01
Show/Entry status for students

kayladavis:
When the entry period is over:
-  the "submission date" changes to "Not accepting Submissions"
- a label is put on once it's in the judging period
- submisssion button removed

once judging period is over:
Label update to show invited status

Example of shows in different states
image

also solves #184

### DIFF
--- a/frontend/src/Student/components/ShowCard.js
+++ b/frontend/src/Student/components/ShowCard.js
@@ -7,7 +7,7 @@ import FaPlusCircle from 'react-icons/lib/fa/plus-circle'
 import FaBook from 'babel-loader!react-icons/fa/book'
 import FaYouTube from 'babel-loader!react-icons/fa/youtube'
 import FaVimeo from 'babel-loader!react-icons/fa/vimeo'
-import { Row, Col } from 'reactstrap'
+import { Row, Col, Alert } from 'reactstrap'
 import moment from 'moment'
 
 const Card = styled.div`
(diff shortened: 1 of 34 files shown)


frontend/src/Student/components/ShowCard.js
```diff
@@ -18,86 +19,160 @@ const Card = styled.div`
...
+          <h5>
+            {props.show.entries.length}/{props.show.entryCap} Submissions
+          </h5>
         </div>
         <div>
-          Accepting Submissions Until: <Moment format='MMMM Do YYYY'>{props.show.entryEnd}</Moment>
+          {moment().isAfter(moment(props.show.entryEnd)) ? (
+            'Submissions are closed'
+          ) : (
+            <AcceptingSubmissionMessage {...props} />
```
billdybas: I don't think you need to break this out into a separate component. Consider moving the ternary up one level in the DOM tree, and return the parent `<div>` from both "components".

frontend/src/Student/components/ShowCard.js
```diff
@@ -18,86 +19,160 @@ const Card = styled.div`
...
         <div>
-          <h5>{props.show.entries.length}/{props.show.entryCap} Submissions</h5>
+          <h5>
+            {props.show.entries.length}/{props.show.entryCap} Submissions
+          </h5>
         </div>
         <div>
-          Accepting Submissions Until: <Moment format='MMMM Do YYYY'>{props.show.entryEnd}</Moment>
+          {moment().isAfter(moment(props.show.entryEnd)) ? (
+            'Submissions are closed'
```
billdybas: "Submissions Closed" or "No Longer Accepting Submissions"

frontend/src/Student/components/ShowCard.js
```diff
@@ -18,86 +19,160 @@ const Card = styled.div`
...
-            md='3'
-            className='text-center align-self-center d-flex justify-content-center align-items-center'
-            style={{minHeight: '10em'}}
-            title={entry.title}
-            key={entry.id}>
-            <EntryThumb entry={entry} />
-          </Col>
-        ))
-      }
+      {moment().isAfter(moment(props.show.entryEnd)) ? (
```
billdybas: How about you always render `<SubmittedEntries>` and conditionally add `<NewSubmission>`

frontend/src/Student/components/ShowCard.js
```diff
@@ -18,86 +19,160 @@ const Card = styled.div`
...
+const EntryContainer = styled.div``
+const JudgingPhase = styled.div`
+  position: relative;
+  border: 1px solid transparent;
+  border-radius: 0.25rem;
+  color: #856404;
+  background-color: #fff3cd;
+  border-color: #ffeeba;
+`
+const Accepted = styled.div`
```
billdybas: I would make all of these labels the same width – in the image you attached, make them all the size of `JudgingPhase`, or make them all be 100% width of their container (or in the case of photos, same width as the photo – which I think is 100% of the container)

kayladavis: Width is now the width of the col

image

billdybas:
I'd rebase off of master before pushing.

PR #203, opened 2018-03-02
Enforce Submission Limit

robmcl4:
Adds both frontend and backend enforcement for the submission limit cap.

On the frontend, the option to submit alone is disabled and a help-tooltip appears to explain why. See photos below.

Disabled 'No' option:

group_submit

And the tooltip:

group_submit_1

On the backend a UserError is thrown with text 'Individual submission limit reached'

### DIFF
--- a/backend/resolvers/mutations/entry.js
+++ b/backend/resolvers/mutations/entry.js
@@ -6,6 +6,7 @@ import Image from '../../models/image'
 import Video from '../../models/video'
 import Other from '../../models/other'
 import Group from '../../models/group'
+import Show from '../../models/show'
 import { ADMIN, IMAGE_ENTRY, OTHER_ENTRY, VIDEO_ENTRY } from '../../constants'
 import { allowedToSubmit, parseVideo } from '../../helpers/submission'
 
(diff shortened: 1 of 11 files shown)


backend/resolvers/mutations/entry.js
```diff
@@ -42,22 +43,49 @@ const createEntry = (entry, entryType, entryId, t) => {
     })
 }
 
+// Rejects the promise if the supplied args indicate the student is doing a
+// single (non-group) submission, but they have met their limit.
+const ensureCanMakeMoreSingleEntries = (
```
kayladavis: Nit: I'm being picky here, but I don't think 'ensure' needs to be part of the function name

frontend/src/Student/components/VideoSubmissionForm.js
```diff
@@ -187,9 +203,27 @@ class VideoSubmissionForm extends Component {
                           name='submittingAsGroup'
                           value='no'
                           required
+                          disabled={!canSubmitAsSingle}
                           checked={values.submittingAsGroup === 'no'}
                         />
-                        <span className='ml-2'>No</span>
+                        {canSubmitAsSingle ? (
```
kayladavis: Can we pull the tooltip out into a higher level component so we're not doing the same thing in three places?

Issue #167, opened 2018-02-15
As an admin, I want to make sure judging is fair across all submissions


billdybas:
Make shuffle algorithm based on #135

We decided to _not_ group submissions by type – just include them in the shuffle order.

billdybas:
May be a dupe of #27

robmcl4:
This might help, using md5.js:

```
/**
 * Performs a repeatable shuffle
 * @param {string} key the shuffle's unique keying which determines order
 * @param {[any]} arr the array to shuffle
 * @param {Function} identifier function that extracts the ID from an array element
 * @returns {[any]} the shuffled array
 */
export function repeatableShuffle (key, arr, identifier) {
  // make a copy of the array so the original is unmodified
  arr = arr.slice()
  // first we sort everything by its identifier, to ensure stability of
  // following shuffle
  arr.sort((a, b) => {
    const idA = identifier(a)
    const idB = identifier(b)
    return idA > idB ? 1 : idA === idB ? 0 : -1
  })
  // modified knuth-shuffle
  for (let i=0; i<arr.length-1; i++) {
    const hash = new MD5()
      .update(`${key}${identifier(arr[i])}`)
      .digest()
    // use hash to generate 'j' such that i <= j < NUM_ELEMS
    const range = arr.length - i - 1
    const j = i + (hash.readUInt32LE(0) % range)
    // swap elems[i], elems[j]
    const tmp = arr[i]
    arr[i] = arr[j]
    arr[j] = tmp
  }
}
```

PR #205, opened 2018-03-03
load entries into the frontend for voting, with unique shuffle applied

robmcl4:
~~DO NOT MERGE~~

ok to merge


frontend/src/Judge/reducers.js
```diff
@@ -50,6 +54,19 @@ const queue = (state = initialQueueState, action) => {
...
 
   switch (action.type) {
+    case actions.FETCH_SUBMISSIONS:
+      const {
+        submissions,
+        username
+      } = action.payload
+      // Merge in the new submissions and perform a shuffle
+      const submissionIds = submissions.map(submission => submission.id)
+      const mergedOrder = Array.from(new Set([...order, ...submissionIds]))
```
billdybas: Can you explain the purpose of `mergedOrder`? Why do we need to account for `...order`?

frontend/src/Judge/reducers.js
```diff
@@ -78,6 +95,34 @@ const queue = (state = initialQueueState, action) => {
...
   // Proxy all queue actions to the particular queue we want to target
   switch (action.type) {
+    case actions.FETCH_SUBMISSIONS:
+      const {
+        submissions,
+        username
+      } = action.payload
+      // First, gather the submissions by show ID (they _should_ be all the same
+      // ID anyway, but just in case)
+      const submissionsByShowId = submissions.reduce(
```
billdybas: In what scenario would they not all have the same ID?

frontend/src/Judge/reducers.js
```diff
@@ -78,6 +95,34 @@ const queue = (state = initialQueueState, action) => {
...
+            submission
+          ]
+        }),
+        {}
+      )
+      // proxy the action to each subqueue
+      return Object.entries(submissionsByShowId).reduce(
+        (accum, [showId, submissions]) => ({
+          ...accum,
+          [showId]: queue(
```
billdybas: This section is scoped to one show, right? 

By that I mean, if we view show 1, will only show 1's data be fetched? Then if we view show 2, will only show 2's data be fetched, not fetching both show 1 and show 2's data (i.e. as we look at more shows, not linearly fetching / shuffling more data)?

frontend/src/Judge/pages/Vote.js
```diff
@@ -68,12 +68,13 @@ class Vote extends Component {
     }).isRequired,
     handlePrevious: PropTypes.func.isRequired,
     handleNext: PropTypes.func.isRequired,
+    fetchSubmissions: PropTypes.func.isRequired,
     submission: PropTypes.object,
     previous: PropTypes.shape({
-      id: PropTypes.number
+      id: PropTypes.string
```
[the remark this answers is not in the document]
billdybas: Ah, I see, Apollo gives them to you as strings – I'll update the example state in the reducer to reflect the change.

frontend/src/Judge/pages/Vote.js
```diff
@@ -150,25 +152,25 @@ class Vote extends Component {
...
-  const { order, viewing } = state.judge.queues[showId]
+  const { order = [], viewing = null } = state.judge.queues[showId] || {}
   const submissions = state.judge.submissions
 
   const obj = {
     show: {
       id: showId
     },
-    submission: submissions[order[viewing]]
+    submission: viewing !== null ? submissions[order[viewing]] : null
```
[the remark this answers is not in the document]
billdybas: Scratch this. `viewing` can be `0` which is falsey, so the first one will never show.

frontend/src/Judge/pages/Vote.js
```diff
@@ -150,25 +152,25 @@ class Vote extends Component {
...
   }
 
   // Show the previous button
-  if (viewing > 0) {
+  if (viewing != null && viewing > 0) {
```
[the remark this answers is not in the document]
robmcl4: Probably `!==` is better, yeah.

frontend/src/Judge/reducers.js
robmcl4: So.... I'm a bit uncertain as to whether this action is pulling in _all_ of the entries or just some of them. I wrote it assuming the latter, but yeah if it's the former then this isn't necessary.

frontend/src/Judge/reducers.js
robmcl4: I just felt a bit odd using `const showId = submissions[0].show.id` and assuming they're all the same. If that's okay though, then I can remove this bit.

frontend/src/Judge/reducers.js
robmcl4: should be.. as above, I can totally modify this so that it assumes you'd only be fetching for one show at once, to clean it up a bit

frontend/src/Judge/pages/Vote.js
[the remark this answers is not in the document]
robmcl4: Yeah `!==` seems better

frontend/src/Judge/actions.js
```diff
@@ -21,19 +23,35 @@ export const fetchSubmission = submissionId => (dispatch, getState, client) => {
 }
 
 export const fetchSubmissions = showId => (dispatch, getState, client) => {
+  // const {shared: {auth: {username}}} = getState()
+  const username = 'me'
```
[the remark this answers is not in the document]
robmcl4: yep, whoops

robmcl4:
Should resolve #167

frontend/src/utils/index.js
```diff
@@ -15,3 +17,36 @@ export const createLoading = namespace => type => ({
...
+  arr = arr.slice()
+  // first we sort everything by its identifier, to ensure stability of
+  // following shuffle
+  arr.sort((a, b) => {
+    const idA = identifier(a)
+    const idB = identifier(b)
+    return idA > idB ? 1 : idA === idB ? 0 : -1
+  })
+  // modified knuth-shuffle
+  for (let i=0; i<arr.length-1; i++) {
```
chtinahow: There's probably a more elegant way to do this, but it looks fine otherwise

frontend/src/Judge/reducers.js
billdybas: The `Submissions` query takes a required `showId` param, and we call `fetchSubmissions()` when we load a particular show to vote on, so theoretically, the entries should have the same `id`

frontend/src/Judge/reducers.js
```diff
@@ -78,6 +90,19 @@ const queue = (state = initialQueueState, action) => {
...
+        return state
+      }
+      // proxy the action to the subqueue
+      const showId = action.payload.submissions[0].show.id
+      return {
+        [showId]: queue(
+          state[showId],
+          action
+        ),
+        ...state
```
billdybas: One last nit: I believe this needs to go before, otherwise you could be overwriting the `[showId]` object you just made:

```js
return {
  ...state,
  [showId]: queue(
    state[showId],
    action
  )
}       
```

kayladavis:
I'm thinking this looks good too. I think after looking at this how I retrieve votes will be different, but that's my PR's problem. Not yours.

PR #207, opened 2018-03-05
 Add backend logic for only looking submissions in start date

chtinahow:
Resolves issue #202 
image
All shows that the student can see will be shown except the ones before the entry start date as shown below in mysql database:
image The Not open show now does not show.


frontend/src/Student/pages/Dashboard.js
```diff
@@ -7,7 +7,7 @@ const Dashboard = () => (
   <Container>
     <Row>
       <Col>
-        <Shows />
+        <Shows studentUsername='user3'/>
```
kayladavis: I think you need to remove the username here.

frontend/src/Student/queries/showForSubmission.graphql
```diff
@@ -0,0 +1,14 @@
+query Show($id: ID!) {
+  show(id: $id) {
+    id
+    name
+    entryCap
```
kayladavis: This entry cap and comment look like their part of what Rob was doing in his PR? Do you mean to pull this info down here?

backend/resolvers/queries/showQuery.js
```diff
@@ -3,43 +3,36 @@ import User from '../../models/user'
...
 }
 
 export function shows (_, args, req) {
-  // If only order is desired, show all shows ordered
-  if (args.orderBy && !args.studentUsername) {
-    return Show.findAll({
-      order: [[args.orderBy.sort, args.orderBy.direction]]
-    })
+  // Students can only look at their own shows
+  console.log(args.studentUsername, req.auth.type)
```
kayladavis: You should remove logging statements

frontend/src/Student/pages/Dashboard.js
chtinahow: Oops, Good catch

backend/resolvers/queries/showQuery.js
chtinahow: Thanks for the catch!

frontend/src/Student/queries/showForSubmission.graphql
chtinahow: what do you mean?

kayladavis: I believe the updated shows.graphql query has what you're asking for. Why have you brought in this file? I don't see it used anywhere in this PR.

chtinahow: Unsure. I will get rid of it

backend/resolvers/queries/showQuery.js
```diff
@@ -3,43 +3,35 @@ import User from '../../models/user'
...
-          } else { // Order shows
-            return Show.findAll({
-              where: {
-                id: showIds
-              },
-              order: [[args.orderBy.sort, args.orderBy.direction]]
-            })
-          }
+          const whereShowIds = { where: { id: showIds } }
+          return Show.findAll(Object.assign({}, whereShowIds, order, entryStart))
```
kayladavis: This is mostly a flaw in the original design, but we also need to include shows that are active for students (date.now is between entry start and entry end) otherwise there's no way for a student to add to a show.

frontend/src/Student/pages/Dashboard.js
```diff
@@ -1,17 +1,22 @@
...
 
 import Shows from '../containers/Shows'
 
-const Dashboard = () => (
+const Dashboard = props => (
   <Container>
     <Row>
       <Col>
-        <Shows />
+        <Shows studentUsername={props.studentUsername}/>
```
kayladavis: When there are no past/current shows we should probably add a simple message like "No shows are currently running, come back later"

billdybas: I'm not sure why you need this change here. In `containers/Shows.js` the `studentUsername` prop is already obtained in `mapStateToProps`. Passing the prop in again (and redefining the same `mapStateToProps`here ) seems redundant.

chtinahow: When I tried removing it, the property was empty when it attempted the query

chtinahow: Done :)

billdybas: Let me test this out before you merge it in – that seems odd.

PR #208, opened 2018-03-07
Adding left and right arrow key input for vote navigation

kayladavis:
Left and right arrow keys can be used to navigate between pages, other key input is ignored. 

Fronend linting isn't running on push. I'm going to try to fix that. Just look at `frontend/src/Judge/pages/Vote.js` for the changes.

In order to test this PR, you can set some default state through your `frontend/src/Judge/reducers.js` file (just like in Bill's last PR):
Change:
```js
const submissions = (state = {}, action) => {
```
to:
```js
const submissions = (state = { 1: {id: '1'}, 2: {id: '2'}, 3: {id: '3'} }, action) => {
```
and
```js
const queues = (state = {}, action) => {
```
to
```js
const queues = (state = { 1: { order: [1, 2, 3], viewing: 0 } }, action) => {
```

### DIFF
--- a/frontend/src/Judge/pages/Vote.js
+++ b/frontend/src/Judge/pages/Vote.js
@@ -117,11 +117,7 @@ class Vote extends Component {
     } = this.props
 
     return (
-      <Container
-        fluid
-        // tabIndex='0'
-        // onKeyDown={(e) => this.handleKeyInput(e, handleNext, handlePrevious, show, previous, next)}
-      >
+      <Container fluid>
         <Row>
           <Col xs='1'>
             {previous && previous.id ? (
--- a/frontend/src/Judge/pages/Vote.js
+++ b/frontend/src/Judge/pages/Vote.js
@@ -40,9 +40,16 @@ const SubmissionContainer = styled.section`
 `
 
 class Vote extends Component {
-  handleKeyInput = (e) => {
-    const { show, handleNext, handlePrevious, submission, previous, next } = this.props
-    if (e.key === "ArrowRight"){
+  handleKeyInput = e => {
+    const {
+      show,
+      handleNext,
+      handlePrevious,
+      submission,
+      previous,
+      next
+    } = this.props
+    if (e.key === 'ArrowRight') {
       if (next && next.id) {
         handleNext()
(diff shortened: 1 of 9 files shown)


frontend/src/Judge/reducers.js
```diff
@@ -7,7 +7,10 @@ import * as actions from './actions'
 //   1: {id: '1', title...},
 //   2: {id: '2', title...}
 // }
-const submissions = (state = {}, action) => {
+const submissions = (
```
billdybas: Looks like this got committed

frontend/src/Student/components/ShowCard.js
```diff
@@ -160,15 +160,20 @@ const ShowCard = props => (
         {moment().isAfter(moment(props.show.entryEnd)) ? (
           <div>No Longer Accepting Submissions</div>
         ) : (
-            <div> Accepting Submissions Until:{' '}
-              <Moment format='MMMM Do YYYY'>{props.show.entryEnd}</Moment></div>
-          )}
+          <div>
+            {' '}
```
billdybas: I realize this has nothing to do with this PR, but this looks like an extraneous space.

frontend/src/Judge/pages/Vote.js
```diff
@@ -40,6 +40,28 @@ const SubmissionContainer = styled.section`
...
+  handleKeyInput = e => {
+    const {
+      show,
+      handleNext,
+      handlePrevious,
+      submission,
+      previous,
+      next
+    } = this.props
+    if (e.key === 'ArrowRight') {
```
billdybas: I was going to comment this:

It looks like in IE and Edge have non-standard key event names for arrow keys.

See: 
- https://caniuse.com/#feat=keyboardevent-key
- https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/8860571/
- https://github.com/shvaikalesh/shim-keyboard-event-key

but then I remembered that React dispatches Synthetic Events, so we don't have to worry about that. (See: [React's source](https://github.com/facebook/react/blob/master/packages/react-dom/src/events/getEventKey.js))

PR #220, opened 2018-03-14
make tests repeatable again

robmcl4:
fixes multiple asynchronous oddities that caused some tests to fail intermittently


billdybas:
LGTM

kayladavis:
lgtm :shipit:

PR #221, opened 2018-03-14
Add Error Handling & Replace Icons

chtinahow:
Please Do not merge


billdybas:
So, I said to take _inspiration_ from the `OneRepoToRuleThemAll`, not to just copy the whole thing!

Mainly because I know from working on the SSE website that its `Status` component isn't the most robust. And since the SSE is connecting to a REST API, not a GraphQL API, the `loading` paradigm is a bit different, especially w/ Apollo.

I'll list some things I'd change here w/ this particular code, but I'd suggest spending some time researching how others suggest doing status messages / spinners w/ Apollo before diving into code. Share what you find in Slack and we can decide on an approach there. Also look into React's Error Boundary components.

Opinions:
- `components/Status.js` is one giant `do` expression, which is still crazy experimental syntax – I'd prefer not to use it here.
- I noticed you copied from `devlop` which has a bug in it where the spinner shows forever if there's an error – though it's fixed on an in-progress PR
- the reducer assumes that you're using `createAction` and `createLoading` from `utils` which we're not using anywhere because we fetch data from GraphQL – we get `loading` from Apollo
- the way of searching for loading messages in the Redux state is brittle
- if you're using font-awesome icons, prefer to use the icon components from `react-icons` instead of using `fa` classes
- we'd like to also show error messages for frontend errors (preventing the app from crashing), something this component doesn't do

robmcl4:
I'm wondering why `package-lock.json` changed. What node version are you running?

billdybas:
This is still WIP. 

I converted all our `react-icons` uses into using font awesome directly because `react-icons` is actually still on v4 and doesn't seem to support animated spinners.

Also moved error handling actions, reducers, etc. into `shared` b/c everyone's going to need it eventually.

TODO:
- [ ] Finish adding error handling to Admin app
  - [ ] Anywhere we fetch data (either through `graphql()` or a Redux action, handle an error case (a good way to test this is to have the backend always throw an error)
- [ ] Add error handling to Judge app
- [ ] Add error handling to Student app
- [ ] Figure out what to do with `apollo-link-error`
- [ ] Add the loading indicator to more places, esp. when we fetch through Redux (will have to make more actions and reducers)

kayladavis:
As far as I can tell apollo-link-errors is mostly usable just for logging right now.

robmcl4:
Is this still WIP? Do I have any action to take here?

kayladavis:
I don't think the loading indicators are everywhere they need to be but that should be fine. I'll change it from WIP. @robmcl4

PR #222, opened 2018-03-16
Adding voting panel

kayladavis:
This PR adds: 
- the voting component
- a success alert
- hooks the vote fetch and mutation into the backend 

Example of just clicked vote:
image


frontend/src/Judge/mutations/sendVote.graphql
```diff
@@ -0,0 +1,5 @@
+mutation sendVote($input: VoteInput!) {
+  vote(input: $input) {
+    value
+  }
+}
```
chtinahow: New line

chtinahow:
Is the success modal supposed to go away after it shows?

backend/resolvers/queries/voteQuery.js
```diff
@@ -6,7 +6,7 @@ import { ADMIN, JUDGE } from '../../constants'
 
 export function votes (_, args, req) {
   const isRequestingOwnJudgeUser = req.auth.username !== undefined &&
-    req.auth.type === JUDGE && req.auth.username === args.judgeUsername
+  req.auth.type === JUDGE && req.auth.username === args.judgeUsername
```
robmcl4: hmm, this spacing seems wrong. Just run the linter eventually plz

frontend/src/Judge/containers/VotePanel.js
```diff
@@ -0,0 +1,38 @@
...
+
+import VotePanel from '../components/VotePanel'
+import SendVote from '../mutations/sendVote.graphql'
+import GetVote from '../queries/entryVote.graphql'
+
+const withMutations = compose(
+  graphql(SendVote, {
+    props: ({ mutate, ownProps }) => ({
+      vote: (value) => mutate({
+        variables: {
```
robmcl4: Hmm, I think the `refetchQueries` thing might have some use here... see [`frontend/src/Admin/containers/AssignJudgesTable.js`](https://github.com/robmcl4/A-B-S-T-R-A-C-T-I-O-N/blob/master/frontend/src/Admin/containers/AssignJudgesTable.js#L58) for how we used it previously. I'd like to hear Bill's opinion, since he put that there in the first place... i'm not too sure what it does exactly.

frontend/src/Judge/components/VotePanel.js
```diff
@@ -0,0 +1,60 @@
...
+  render() {
+    const {
+      vote,
+      data
+    } = this.props
+    const entryVote = data.vote
+    return (
+      <div>
+        <ButtonGroup >
+          <VoteSpacing><Button color="primary" size="lg" active={entryVote && entryVote.value === 0} onClick={() => vote(0).then(this.handleVote())}>No</Button></VoteSpacing>
```
robmcl4: Instead of the `<VoteSpacing>` block, could we just add 10px of margin to the `<Button>`? I _think_ this has the added advantage of an extended click-target... Bill managed to extend the click target area for the navigation buttons, but i'm not certain what method is appropriate here. I think it's OK to merge if the CSS gets too confusing though.

kayladavis:
@chtinahow It currently dismisses after 3 seconds. I couldn't decide the best way to handle it. I tried hooking into clicks, but that fails if they try to use the keyboard method anyway. I could pull in a toast package that shows when the alert/toast/whatever will time out.

As for the "know what they chose" bit, the button should stay highlighted.

frontend/src/Judge/containers/VotePanel.js
billdybas: Apollo's mutations send and forget. You can specify `refetchQueries` to get data back that may have changed because of your mutation. It's not really the direction Apollo steers you in – they'd rather have you update the local cache rather than make another network request – but because it was easier to implement / reason about and our app is small, I used `refetchQueries` anyways.

We'll need it here to display what choice a Judge made, and again, it's probably easier to use `refetchQueries` than to do cache updates, though I would think this would be more of a candidate for cache updates than the Admin side (since votes are scoped to a particular Judge).

PR #225, opened 2018-03-20
Adding videos to Voting page

kayladavis:
This PR renders vimeo and youtube videos.
Example:
image

I still recommend testing this out yourself, as a screenshot can't capture all controls.

### DIFF
--- a/frontend/config/apollo.js
+++ b/frontend/config/apollo.js
@@ -1,6 +1,7 @@
 import { ApolloClient } from 'apollo-client'
 import { HttpLink, InMemoryCache } from 'apollo-client-preset'
 import { ApolloLink, concat } from 'apollo-link'
+import { IntrospectionFragmentMatcher } from 'apollo-cache-inmemory'
 
 const httpLink = new HttpLink({uri: 'http://localhost:3000/graphql'}) // TODO: Read from .env
 
@@ -16,6 +17,30 @@ const authMiddleware = new ApolloLink((operation, forward) => {
   return forward(operation)
 })
 
+const fragmentMatcher = new IntrospectionFragmentMatcher({
+  introspectionQueryResultData: {
+    __schema: {
+      types: [ 
+        {
+          "kind": "INTERFACE",
+          "name": "Entry",
+          "possibleTypes": [
+            {
+              "name": "Photo"
+            },
+            {
+              "name": "Video"
+            },
+            {
+              "name": "OtherMedia"
+            }
+          ]
+        }
+      ] 
+    }
+  }
+})
+
 const client = new ApolloClient({
   link: concat(authMiddleware, httpLink),
   cache: new InMemoryCache({
@@ -27,7 +52,8 @@ const client = new ApolloClient({
         default:
           return `${o.__typename}:${o.id}`
       }
-    }
+    },
+    fragmentMatcher: fragmentMatcher
   })
 })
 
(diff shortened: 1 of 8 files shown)


frontend/config/apollo.js
```diff
@@ -16,6 +17,30 @@ const authMiddleware = new ApolloLink((operation, forward) => {
   return forward(operation)
 })
 
+const fragmentMatcher = new IntrospectionFragmentMatcher({
```
robmcl4: neat!

frontend/src/Judge/components/VideoSubmission.js
```diff
@@ -1,5 +1,23 @@
...
+import ReactPlayer from 'react-player'
+import styled from 'styled-components'
 
-const VideoSubmission = props => <div>Video Submission</div>
+// Video Providers
+const youtubeBaseURL = 'https://www.youtube.com/watch?v='
+const vimeoBaseURL = 'https://vimeo.com/'
+
+const videoStyle = {
+  minHeight: '500px',
```
robmcl4: nit: trailing comma

frontend/src/Judge/queries/submissions.graphql
```diff
@@ -1,8 +1,20 @@
 query Submissions($id: ID!) {
   submissions: entries(showId: $id) {
-    id,
+    id
+    title
+    entryType
     show {
       id
     }
+    ... on Video {
```
robmcl4: I like this syntax ... could you possibly also convert `frontend/src/Student/queries/shows.graphql` to use it as well? If you don't have time that's fine, don't let it block the merge.

frontend/src/Judge/components/VideoSubmission.js
```diff
@@ -1,5 +1,23 @@
...
+}
+
+const VideoSubmission = props => (
```
billdybas: Here's the component w/ the centering fix you asked me about yesterday – renders the player at the default size and prevents the player from having black bars.

```jsx
import React from 'react'
import PropTypes from 'prop-types'
import ReactPlayer from 'react-player'

// Video Providers
const YOUTUBE_BASE_URL = 'https://www.youtube.com/watch?v='
const VIMEO_BASE_URL = 'https://vimeo.com/'

const PlayerContainer = styled.div`
  margin-top: 50px;

  /* Center the top-most 'div' that ReactPlayer wraps around the 'iframe'  */
  & > div {
    margin: 0 auto;
  }
`

const VideoSubmission = props => (
  <PlayerContainer>
    {props.provider === 'youtube' ? (
      <ReactPlayer url={`${YOUTUBE_BASE_URL}${props.videoId}`} playing controls />
    ) : props.provider === 'vimeo' ? (
      <ReactPlayer url={`${VIMEO_BASE_URL}${props.videoId}`} playing controls />
    ) : null}
  </PlayerContainer>
)

VideoSubmission.propTypes = {
  provider: PropTypes.string.isRequired,
  videoId: PropTypes.string.isRequired
}

export default VideoSubmission

```

billdybas: Also, with your styles and my styles, when you vote on a video submission, the voting panel shifts down on the page when the alert appears. You should hunt down what's causing that and prevent the voting panel from moving.

billdybas: I do like how your styles make the video responsive. The 500px min-height, however, feels a bit tight on my laptop (full screen window), and not a fan of the black bars:

screen shot 2018-03-20 at 8 26 03 am

billdybas:
LGTM.

If you wanted to change the outlined button, I think you'd need to conditionally set `background-color: white; opacity: 1;` You can also do nested selectors in styled-components – you could convert the top level `div` in `VotePanel` to a styled component and then target the outlined button class w/ those styles.

PR #227, opened 2018-03-20
make judging urls refreshable

robmcl4:
Adds refreshable URLs. This is accomplished by removing the 'viewing' property from state and instead relying on conneccted-react-router's state.

### DIFF
--- a/frontend/src/Judge/pages/Vote.js
+++ b/frontend/src/Judge/pages/Vote.js
@@ -89,19 +89,6 @@ class Vote extends Component {
     document.addEventListener('keydown', this.handleKeyInput)
   }
 
-  componentWillReceiveProps (nextProps) {
-    // If we receive props where 'viewing' exists but the url ?on={int} doesn't,
-    // set the latter to the former
-    if (nextProps.submission !== null && !queryString.parse(nextProps.location.search).on) {
-      const thing = {
-        ...queryString.parse(nextProps.location.search),
-        on: nextProps.submission.id
-      }
-      const newQueryString = queryString.stringify(thing)
-      this.props.history.replace(`/show/${nextProps.show.id}/vote?${newQueryString}`)
-    }
-  }
-
   render () {
     const { setViewing, submission, previous, next, vote } = this.props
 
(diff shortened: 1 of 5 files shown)


billdybas:
I ran into some bugs:

1. Invalid query params break. Eg. `on=a` shows me a blank screen. It should show me the first entry.

2. If the URL is `/vote` and I vote on something, my choice doesn't highlight. Additionally, if that vote was the last vote, my page updates to the first submission.

frontend/src/Judge/pages/Vote.js
```diff
@@ -164,16 +127,57 @@ class Vote extends Component {
 
 const mapStateToProps = (state, ownProps) => {
   const showId = ownProps.match.params.id
-  const { order = [], viewing = null } = state.judge.queues[showId] || {}
+  const { order = [], loadingVotes = true, loadingSubmissions = true } =
+    state.judge.queues[showId] || {}
+  let { on: submissionId } = queryString.parse(state.router.location.search)
+
+  // If this submissionId is not parsable as a base-10 integer, throw it out
+  if (isNaN(parseInt(submissionId, 10))) {
```
billdybas: This check is not needed, as invalid id's won't be found in `order` by the check below.

Additionally, having this check couples the frontend to the id mechanism on the backend - if we changed to a public facing random string like we mentioned, this would break.

frontend/src/Judge/reducers.js
```diff
@@ -89,8 +88,8 @@ const queue = (state = initialQueueState, action) => {
 // Each show has a queue whose key in the queues state is the show's id
 // Example State:
 // {
-//   1: {order: [1, 2, 3], viewing: 0},
-//   2: {order: [9, 4, 5, 8, 6, 7], viewing: 5}
+//   1: {order: [1, 2, 3], loadingSubmissions: true, loadingVotes: false},
+//   2: {order: [9, 4, 5, 8, 6, 7], loadingSubmissions: true, loadingVotes: false}
```
billdybas: Also noticed the values in `order` in these comments should be made into strings – put single quotes around them.

PR #228, opened 2018-03-21
make an admin's show resolver show all

### LINKED ISSUE #224: Admins Can't See Current Shows When "Viewing As Judge"
Admin's can't assign themselves to a show, so the current queries for shows on the dashboard won't return anything for Admins when they're viewing as a Judge. Admins should have the same experience as Judges with the caveat that they're "assigned" to all shows.

robmcl4:
Makes an admin's User.shows resolver indicate that they are assigned to all shows -- resolves #224.


robmcl4:
Good catch. Fixed in 155c50b

PR #229, opened 2018-03-21
Adding Photo voting view

kayladavis:
This PR add the ability to view photos on the judging page. They are rendered to max fit in the voting container without stretching. When clicked on a "lightbox" feature is rendered. This renders the image max size for the entire window.

I also changed the video rendering to fill the component
Examples: 
image
image
image

### DIFF
--- a/frontend/src/Judge/components/OtherMediaSubmission.js
+++ b/frontend/src/Judge/components/OtherMediaSubmission.js
@@ -31,24 +31,26 @@ class OtherMediaSubmission extends Component {
   static propTypes = {
     path: PropTypes.string.isRequired
   }
-  render() {
+  render () {
     const { path } = this.props
 
     return (
       <Fragment>
-        <Alert
-          color='warning'
-          style={{ textAlign: 'center' }}
-        >
+        <Alert color='warning' style={{ textAlign: 'center' }}>
           This submission may need to be viewed in person
         </Alert>
         {path.match('.pdf$') ? (
           // TODO make this URL responsive to deploy environment
-          <PdfWell href={`//localhost:3000/static/uploads/${path}`} target='_blank'>
-            <PdfImg style={{backgroundImage: `url(${pdf})`}}  />
+          <PdfWell
+            href={`//localhost:3000/static/uploads/${path}`}
+            target='_blank'
+          >
+            <PdfImg style={{ backgroundImage: `url(${pdf})` }} />
             <h3>Click to View</h3>
           </PdfWell>
-        ) : (<PhotoSubmission path={path} />)}
+        ) : (
+          <PhotoSubmission path={path} />
+        )}
       </Fragment>
     )
   }
--- a/frontend/src/Judge/components/OtherMediaSubmission.js
+++ b/frontend/src/Judge/components/OtherMediaSubmission.js
@@ -26,8 +26,6 @@ const PdfWell = styled.a`
   background-color: #666666;
   box-shadow: inset 2px 2px 1px rgba(0, 0, 0, 0.05);
 `
-const PdfText = styled.div`
-`
 
 class OtherMediaSubmission extends Component {
   static propTypes = {
@@ -48,9 +46,7 @@ class OtherMediaSubmission extends Component {
           // TODO make this URL responsive to deploy environment
           <PdfWell href={`//localhost:3000/static/uploads/${path}`} target='_blank'>
             <PdfImg style={{backgroundImage: `url(${pdf})`}}  />
-            <PdfText>
-              <h3>Click to View</h3>
-            </PdfText>
+            <h3>Click to View</h3>
           </PdfWell>
         ) : (<PhotoSubmission path={path} />)}
       </Fragment>
--- a/frontend/src/Judge/components/OtherMediaSubmission.js
+++ b/frontend/src/Judge/components/OtherMediaSubmission.js
@@ -1,5 +1,111 @@
-import React from 'react'
+import React, { Component, Fragment } from 'react'
+import styled from 'styled-components'
+import PropTypes from 'prop-types'
+import Lightbox from 'react-images'
+import { Alert } from 'reactstrap'
 
-const OtherMediaSubmission = props => <div>Other Submission</div>
+import pdf from 'assets/pdf.svg'
+
+const PhotoContainer = styled.img`
+  height: 90%;
+  object-fit: scale-down;
+  width: 100%;
+`
+
+const PdfContainer = styled.img`
+  height: 100%;
+  object-fit: scale-down;
+  width: 45%;
+  background-color: #666666;
+  box-shadow: inset 2px 2px 1px rgba(0, 0, 0, 0.05);
+`
+
+const PdfText = styled.div`
+  position: absolute;
+  top: 80%;
+  left: 50%;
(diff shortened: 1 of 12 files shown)


kayladavis:
The last commit 0b09e51 can be reverted if you don't want it. There's just a lot of overlap with Photos and I had time, so I added it. This is what the othermedia page looks like:
image
or
image

frontend/src/Judge/components/OtherMediaSubmission.js
```diff
@@ -1,5 +1,111 @@
...
+            <Alert
+              color='info'
+              style={{
+                marginTop: '45px',
+                width: '45%',
+                left: '50%',
+                transform: 'translate(-50%, -50%)'
+              }}
+            >
+              This submission may need to be viewed in person
```
chtinahow: You might need some padding around this alert so it doesn't overlay, but this may be a needed thing for the voting panel.

frontend/src/Judge/components/OtherMediaSubmission.js
```diff
@@ -1,5 +1,111 @@
...
+import PropTypes from 'prop-types'
+import Lightbox from 'react-images'
+import { Alert } from 'reactstrap'
 
-const OtherMediaSubmission = props => <div>Other Submission</div>
+import pdf from 'assets/pdf.svg'
+
+const PhotoContainer = styled.img`
+  height: 90%;
+  object-fit: scale-down;
```
billdybas: `object-fit` is not supported in IE 11 which we need to support because of Windows 7: https://caniuse.com/#feat=object-fit

frontend/src/Judge/components/OtherMediaSubmission.js
```diff
@@ -1,5 +1,111 @@
...
+class OtherMediaSubmission extends Component {
+  static propTypes = {
+    path: PropTypes.string.isRequired
+  }
+
+  constructor (props) {
+    super(props)
+
+    this.state = {
+      lightboxIsOpen: false
```
billdybas: `isLightboxOpen`

frontend/src/Judge/components/OtherMediaSubmission.js
```diff
@@ -1,5 +1,111 @@
...
+  openLightbox = () => {
+    this.setState({ lightboxIsOpen: true })
+  }
+
+  render () {
+    const { path } = this.props
+
+    return (
+      <Fragment>
+        {path.endsWith('.pdf') ? (
```
billdybas: `endsWith` is not supported in IE, so unless we get it polyfilled by Babel, it'll need to change: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith

frontend/src/Judge/components/OtherMediaSubmission.js
```diff
@@ -1,5 +1,111 @@
...
+          <div style={{ height: '80%', marginTop: '25px' }}>
+            <a href={`//localhost:3000/static/uploads/${path}`} target='_blank'>
```
billdybas: Probably add a `TODO` to come back and fix this so we don't forget it

frontend/src/Judge/components/OtherMediaSubmission.js
```diff
@@ -1,5 +1,111 @@
...
+              <PdfContainer src={pdf} />
+              <PdfText>
+                <h3>Click to View</h3>
+              </PdfText>
+            </a>
+            <Alert
+              color='info'
```
billdybas: I think we originally agreed we'd use `warning` as the color of this alert.

frontend/src/Judge/components/PhotoSubmission.js
```diff
@@ -1,5 +1,57 @@
...
+
+  openLightbox = () => {
+    this.setState({ lightboxIsOpen: true })
+  }
+
+  render () {
+    const { path } = this.props
+
+    return (
+      <Fragment>
```
billdybas: This seems to be duplicated here and in the `OtherMediaSubmission` component. Consider creating a component which is shared between them.

frontend/src/Judge/components/VideoSubmission.js
```diff
@@ -8,25 +8,68 @@ const YOUTUBE_BASE_URL = 'https://www.youtube.com/watch?v='
...
+    super(props)
+
+    this.state = {
+      height: 0,
+      width: 0
+    }
+    this.updateDimensions = this.updateDimensions.bind(this)
+  }
+
+  updateDimensions () {
```
billdybas: If you change this to `updateDimensions = () => {`, you can remove `this.updateDimensions = this.updateDimensions.bind(this)` above.

frontend/src/Judge/components/VideoSubmission.js
```diff
@@ -8,25 +8,68 @@ const YOUTUBE_BASE_URL = 'https://www.youtube.com/watch?v='
...
+
+  componentDidMount () {
+    this.updateDimensions()
+    window.addEventListener('resize', this.updateDimensions)
+  }
 
-export default VideoSubmission
\ No newline at end of file
+  componentWillUnmount () {
+    window.addEventListener('resize', this.updateDimensions)
```
billdybas: `window.removeEventListener`?

frontend/src/Judge/pages/Vote.js
```diff
@@ -99,12 +104,14 @@ class Vote extends Component {
...
   componentWillUnmount () {
     document.removeEventListener('keydown', this.handleKeyInput)
+    document.body.style = 'background: white'
   }
 
   componentDidMount () {
     this.props.fetchSubmissions()
     this.props.fetchVotes()
     document.addEventListener('keydown', this.handleKeyInput)
+    document.body.style = 'background: #777777'
```
billdybas: Not sure if I like this, feels a bit hacky.

frontend/src/Judge/components/OtherMediaSubmission.js
billdybas: In our original sketches, we had this alert at the top of the screen.

Is it more noticeable at the top or the bottom?

Either way, agree w/ Tina it should be a block element and push down the Vote Panel as necessary. I can help w/ the CSS if necessary.

frontend/src/Judge/components/PhotoSubmission.js
kayladavis: `<PhotoSubmission path={path}` used in OtherMedia

frontend/src/Judge/components/Submission.js
```diff
@@ -11,26 +11,30 @@ const PHOTO = 'PHOTO'
 const OTHER = 'OTHER'
 
 const Submission = props => (
-  <div>
+  <div style={{ height: '100%' }}>
```
chtinahow: So, there's some stuff that may help with the spacing, in this file I did this: 
image which gives the submission some breathing room like this: 
image

kayladavis: This didn't end up working for me. I learned how to use flexbox and fixed it that way.

frontend/src/Judge/pages/Vote.js
kayladavis: I think this is the only way to handle it and account for overscrolling it.

PR #230, opened 2018-03-24
Add thumbnail generation to image upload

robmcl4:
When images are uploaded, thumbnails are also generated. These thumbnails are non-cropped and maintain aspect ratio of the original, but are bounded to 400w x 300h.

Thumbnails are found (for example) at: `e/5/e5beaf80-2fb7-11e8-9ad7-49aad71b7572_thumb.jpg`.

Modifies the ShowCard's photo preview to use the new thumbnail.

Uses [sharp](https://github.com/lovell/sharp) for image processing, no additional binaries needed.


robmcl4:
Also, changes to v4 uuids, to reduce possibility of enumeration

robmcl4:
065cfc4  should correctly scale (and not crop) both landscape _and_ portrait photos now

billdybas:
Do we want square images to be 300 x 300px or 400 x 400px?

robmcl4:
I don't really mind either way, 400 x 400px seems fine.

PR #233, opened 2018-03-25
allow students to query open show calls

### LINKED ISSUE #224: Admins Can't See Current Shows When "Viewing As Judge"

### LINKED ISSUE #226: Students Can't See Shows to Make Submissions
Mentioned in #207. There's currently no way for students to make a submission to an in-progress show (if they don't have any submissions to it yet).

robmcl4:
Modifies the Shows query to show all Shows to which the student has submitted, plus all open shows.

Fixes #226

### DIFF
--- a/backend/resolvers/queries/showQuery.js
+++ b/backend/resolvers/queries/showQuery.js
@@ -22,7 +22,6 @@ export function shows (_, args, req) {
   const order = args.orderBy ? {order: [[args.orderBy.sort, args.orderBy.direction]]} : {}
   // If username given, show all shows the student has submitted to
   if (args.studentUsername) {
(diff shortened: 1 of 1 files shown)


robmcl4:
Hmm.... `DESC` should already sort older shows to the bottom, since dates in the past have lower order. Experimentally, when I make multiple shows (with some in the past) the most current ones seem to be at the top.

billdybas:
Ah, you're right. I interpreted my test data wrong. `DESC` is 👍

PR #234, opened 2018-03-25
Add show-specific page for Admins, with tabs

robmcl4:
Adds `/show/:id?tab={1,2,3}` to the Admin interface, with three tabs 'Details', 'Submissions', and 'Judge'

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -57,6 +57,7 @@
     "react-moment": "^0.7.0",
     "react-redux": "^5.0.6",
     "react-router-dom": "^4.2.2",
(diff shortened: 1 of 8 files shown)


robmcl4:
NOTE: in the story it's specified that we manage tab navigation by path, but I used the query-string

billdybas:
This is way overkill and could be done just with React Router. 

Use this directly or for inspiration: https://github.com/chacestew/react-router-tabs

And structure the URLs how the story says to.

robmcl4:
Rewrote to use react-router-tabs in ed72791

Issue #214, opened 2018-03-12
As an Admin, I want a page where I can view reports and download images for a show


billdybas:
This story involves designing the "Details" tab on the "View Details" show page.

This page should include a section for Reports and a section for Downloads

On the reports section, there should be links / buttons to generate each report we build.
On the downloads section, there should be a button to download the ZIP of a show's submission images

chtinahow:
Download csv link

chtinahow:
Put the buttons in, but not hook the generate reports just yet. Hook up the  download submissions to the button

PR #236, opened 2018-03-26
Add download button

robmcl4:
~DO NOT MERGE~ OK to merge

adds a 'download zip' button, plus other (currently useless) buttons for reports, addresses #214


kayladavis:
Lgtm. To Tina's point,  about the no submissions. I think we shouldn't worry about that edge case at this time. It's not a realistic problem, and we can focus our efforts elsewhere.

frontend/src/shared/actions.js
```diff
@@ -41,3 +42,22 @@ export const switchToAdmin = () => (dispatch, getState, client) => {
     type: SWITCH_TO_ADMIN
   })
 }
+
+export const getDownloadToken = () => (dispatch, getState, client) => {
+  const { shared: { auth: { token } } } = getState()
+
+  // TODO replace this with the actual deployed url
+  return fetch('//localhost:3000/auth/downloadToken', {
```
billdybas: This should use axios. Fetch isn't supported in IE.

PR #237, opened 2018-03-28
Add submissions table

robmcl4:
I expect this might need some fix-up

Photo of basic setup:

The 'dimensions' column is sorted by the _area_

The 'Artist' column sorts by the submitter

submission table


frontend/src/Admin/components/ShowSubmissionsTab.js
```diff
@@ -1,5 +1,238 @@
...
+  }
+
+  onDismiss = () => {
+    this.setState({
+      alertVisible: false,
+      alertType: ''
+    })
+  }
+
+  updateShowInvite = (id, value) => {
```
billdybas: `updateInvitation`

frontend/src/Admin/components/ShowSubmissionsTab.js
```diff
@@ -1,5 +1,238 @@
...
+    const { updateInvite } = this.props
+    updateInvite(id, value)
+      .then(() => {
+        this.setState({
```
billdybas: I don't know if showing alerts is useful here since the star icon will have changed color signaling a change.

frontend/src/Admin/components/ShowSubmissionsTab.js
```diff
@@ -1,5 +1,238 @@
...
+              Header: 'Score',
+              accessor: 'score',
+              Cell: ({ original: submission }) => submission.score.toFixed(3)
+            },
+            {
+              Header: 'Invited',
+              accessor: 'invited',
+              Cell: ({ original: submission }) =>
+                submission.invited ? (
+                  <a
```
billdybas: Don't use an `<a>` tag here. Use something like `<span>`

frontend/src/Admin/components/ShowSubmissionsTab.js
```diff
@@ -1,5 +1,238 @@
...
+                      })
+                    </p>
+                    <p>Participants: {submission.group.participants}</p>
+                  </Fragment>
+                )
+            },
+            {
+              id: 'dimensions',
+              accessor: submission =>
+                // Allows for sorting by area
```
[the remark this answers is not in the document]
kayladavis: lets go with what we have now. We can easily tweak this later.

frontend/src/Admin/components/ShowSubmissionsTab.js
kayladavis: we had the alert in the original story. I say we keep it for now. We can remove it later if Nanette doesn't like it.

billdybas: An alert is not a toast, an imo adds too much clutter to the page. We can keep it in for now and add it to our tech debt list.

frontend/src/Admin/components/ShowSubmissionsTab.js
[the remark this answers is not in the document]
billdybas: She said that sorting by area is good.

Issue #231, opened 2018-03-24
New entries don't appear on the dashboard after submitting them


robmcl4:
When a student submits an entry they are immediately redirected to their dashboard. The new entry is not displayed there, but it should be. I believe this is due to apollo caching.

PR #238, opened 2018-03-28
clear apollo cache after submitting entry

robmcl4:
Clears the Apollo in-memory cache after an Entry is submitted: this causes graphql to refetch the summary Shows data and when redirected back the homepage the student can find their entry listed there.

Addresses #231


robmcl4:
Abandoning this PR: will attempt to instrument refetchQueries instead of undocumented features

PR #242, opened 2018-03-30
Add judge progress loading-bars

robmcl4:
Adds loading-bars which indicate judging progress. See below for photos.

Adds a backend change to add property on entries called 'votes' which exposes all votes made on that entry.

image

### DIFF
--- a/frontend/src/Admin/components/ShowJudgesTab.js
+++ b/frontend/src/Admin/components/ShowJudgesTab.js
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { Fragment } from 'react'
 import PropTypes from 'prop-types'
 import { Row, Col, Progress, Button } from 'reactstrap'
 import { Link } from 'react-router-dom'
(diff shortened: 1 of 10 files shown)


frontend/src/Admin/components/ShowJudgesTab.js
```diff
@@ -1,5 +1,90 @@
...
+          <Row key={judge.username} style={{ marginBottom: '1.5rem' }}>
+            <Col md='3'>
+              {judge.firstName} {judge.lastName} ({judge.username})
+            </Col>
+            <Col>
+              <Progress
+                value={numVotes / numEntries * 100}
+                color={colorForProgress(numVotes / numEntries)}
+              >
+                {numVotes} / {numEntries}
```
billdybas: On Judges who haven't started voting yet, this currently just shows a "/"



Should show 0 / <total>, ideally in the center of the bar in black.

frontend/src/Admin/components/ShowJudgesTab.js
```diff
@@ -1,5 +1,90 @@
...
+  }
+  if (progress < 1) {
+    return null // default color
+  }
+  return 'success'
 }
+
+const ShowJudgesTab = ({ show }) => {
+  // Extract all votes we care about
+  const allVotes = show.entries.reduce(
```
billdybas: This section until `const numEntries` should be moved up into a container. The component should only worry about how the progress bars are rendered, not how the data should be formatted first. Pass down that formatted data as `props`.

frontend/src/Admin/components/ShowJudgesTab.js
```diff
@@ -1,5 +1,90 @@
...
+  return (
+    <div>
+      <Row style={{ marginBottom: '1.5rem' }}>
+        <Col className='text-right'>
+          <Link to={`/show/${show.id}/judges/assign`}>
+            <Button color='primary'>Manage Judges</Button>
+          </Link>
+        </Col>
+      </Row>
+      {show.judges.map(judge => {
```
billdybas: Are Judges guaranteed to be in sorted order by their `username`?

billdybas: If no Judges are assigned to this show, you should render something like: "No judges are assigned to this show. Visit the "Manage Judges" page to assign judges to this show." maybe with a link to the page.

frontend/src/Admin/components/ShowJudgesTab.js
```diff
@@ -1,5 +1,90 @@
...
+  if (progress < 0.25) {
+    return 'warning'
+  }
+  if (progress < 1) {
+    return null // default color
+  }
+  return 'success'
 }
+
+const ShowJudgesTab = ({ show }) => {
```
billdybas: 2 Questions:

1. This doesn't seem to show the voting progress of Admins. Should it?

2. What currently happens on the backend to a Judge's votes if they are unassigned from a show? 
  a) Should the votes be kept or deleted?
  b) Should we disable Judge assignment after judging has started?

frontend/src/Admin/components/ShowJudgesTab.js
robmcl4: Hmm... I think in this case it's easier to just not show it at all. The CSS for the former would be hacky I suspect... since I don't think there's a markup for it

frontend/src/Admin/components/ShowJudgesTab.js
robmcl4: 1. I'm not sure... admins don't get 'assigned' to a show, so that's a bit tricky to instrument.
2. Currently the votes are kept, but I suspect that this shouldn't be the case since it will influence the aggregate score. Option 2 seems nice, but I feel like we should prefer freedom to assign judges last-minute over restriction

frontend/src/Admin/components/ShowJudgesTab.js
robmcl4: hmm... I suppose not

robmcl4:
I'm a little bit stuck deer-in-headlights on how to handle listing admins here: on the one hand, it makes sense. On the other hand, we don't have a backend route to easily list all of the admins. But that could be done. Thoughts?

frontend/src/Admin/components/ShowJudgesTab.js
billdybas: 1. I think it's fine at the moment to not list Admins. The main use case here is for Admins to see which Judges they need to bother to vote. I'm willing to push listing Admins off to "Future Work"

2. I agree with both your points. I think a happy medium would be to delete a Judge's votes if they are unassigned from a show. We could also show a confirmation dialog on the assignment page to make sure Admins know that.

Issue #26, opened 2017-09-19
As a Judge I would like to see my progress on judging individual calls, so that I can see how much time I should set aside to finish judging.


robmcl4:
As a Judge I would like to see my progress on judging individual calls, so that I can see how much time I should set aside to finish judging.

billdybas:
On the show cards on the Judge Dashboard, show a progress indicator (eg. 27 / 159)

Show a similar indicator on the top right of the Judging Interface so Judges know at a glance how much they have left to judge.

chtinahow:
@billdybas The progress indicator should be on the top right according to the interface we came up with
![basic_voting_interface](https://user-images.githubusercontent.com/8568993/36273706-65527d30-1253-11e8-9378-15e10404f8c4.jpg)

chtinahow:
This will just be the number out of how many completed

billdybas:
We determined the most intuitive thing to put here was the index of the entry the Judge is looking at out of the total – gives context on how many are to the left and right.

PR #249, opened 2018-04-06
Add Voting Progress to Voting Page

billdybas:
#26 

I'm not happy with this implementation as it fetches `entries` and `ownVotes` for all a Judge's shows, but this was the only way to currently do it without making backend changes.

If someone wants to add in backend changes to get them for an individual show, do that.


robmcl4:
Wait... I was under the impression that the progress display was just a glorified page number for the entry being viewed, not a count of votes that have been cast. If that is sufficient to meet the requirement, that's way easier...

PR #250, opened 2018-04-08
Formik & Judges Page

billdybas:
#156 

- Converts remaining forms to formik & removes redux-form
- Adds a "Judges" page which lists all judges w/ the "Add Judge" form
- Sorts a show's submissions by score by default
- Standardizes use of recompose


frontend/src/Admin/pages/AssignJudges.js
```diff
@@ -3,6 +3,7 @@ import { Container } from 'reactstrap'
 import styled from 'styled-components'
```
kayladavis: I'm not sure if my test data is in a weird state but when I go to the `judges/assign` page from the judges tab get a white page because of the this error: `Cannot read property '_subRows' of undefined`

frontend/src/Admin/components/CreateShowForm.js
```diff
@@ -1,120 +1,216 @@
...
+                  />
+                  {this.renderErrors(touched, errors, 'entryCap')}
                 </FormGroup>
               </Col>
+              <Col xs={6}>
+                <Row>
+                  <Col>
+                    <FormGroup>
+                      <Label>Submission Dates</Label>
+                      <CalendarContainer>
```
kayladavis: Is there a simple way to allow time input on this form too? When the show starts and ends on the same day it doesn't work (I know this has been an issue, but this seems like a good time to fix it if it's a simple change).

billdybas: Simple – no. Possible – probably.

Yes, the validation on the frontend requires the day after. I think the validation on the backend only requires greater than time. 

We can double check what MySQL defaults the time to (or set it ourselves in our API) and indicate that in a tooltip on this page – we'll have to show it to students, too, so they don't miss the deadline.

frontend/src/Admin/pages/AssignJudges.js
billdybas: That's real strange. Can you give me more information so I can diagnose the problem? Browser & Version? Example Data? 

My first guess is `react-table` is having an issue rendering the data it's provided.

When I try visiting the "assign" page from that tab, everything works swimmingly in latest Chrome, Firefox, and Safari 10.1

kayladavis: I can't repro this on my laptop, this section is good to go.

frontend/src/Admin/components/CreateShowForm.js
kayladavis: I think we need to handle time before we hand this off to Nanette, but not handling it in this change is fine.

PR #251, opened 2018-04-09
 Removes Group Name & Adds Full UTF-8 Support

billdybas:
Also fixes issue w/ ZIP download when files can't be found on the server.

I also modified the naming scheme of the ZIP:
- For Individual: `/Show Name/{Invited, Not Invited}/Lastname, Firstname - Title <(num)>.jpg`
- For Group: `/Show Name/{Invited, Not Invited}/Lastname, Firstname & Participants - Title <(num)>.jpg`

~~~I noticed that for a group submission, we'll get two downloads, one for the submitter and one for the participants. I guess this behavior is alright, but should we dedupe submissions?~~~


backend/routes/download.js
```diff
@@ -210,11 +209,7 @@ router.route('/zips/:showId')
             return submissionsWithSubmitters.reduce((arr, {user, group, entries}) => {
               // ['Clark Kent - Daily Planet Office']
               const newSubmissionSummaries = entries.map(({path, title, invited}) => {
-                const entryNamePrefix = (
-                  (user ? `${user.lastName} ${user.firstName}` : group.name) +
-                  ' - ' +
-                  title
-                )
+                const entryNamePrefix = `${(user ? `${user.lastName}, ${user.firstName}` : group.participants)} - ${title}`
```
kayladavis: I think we definitely need to dedupe the group submissions. When something is a group submission we should just include that one (not a separate one for the creator too).

For group name I think we should do something like this (because I don't think the  group creator is listed in participants): 
`/Show Name/{Invited, Not Invited}/{creatorName}&Participants - Title <(num)>.jpg`

billdybas: @kayladavis Is the deduping problem solved? Do I just need to merge in master?

kayladavis: I think for the purposes Nanette mentioned she uses this for, just the participants is fine. I'm okay will it being what you changed it to.

billdybas: Fixed it so that group creator is included

PR #239, opened 2018-03-28
Small Fixes, Rewording, Visual Quirks

billdybas:
~~~@robmcl4 Add to this PR any of the small things you wanted to change.~~~

This PR includes:
- Rewording the Student Success Modal
- Fixing Caching Issues on Student Dashboard
- Moving Frontend URLs into Constants
- Fixing a Backend Regression
- Reskinning the Voting Interface to blend the buttons into the grey background (i.e. remove the white section)

PR #252, opened 2018-04-09
Adding CSV export to Admin Details tab

kayladavis:
This PR allows an admin to download the information for all entries on a show as a csv file.

To do this, it follows a similar process to the zip download functionality. Here is an example of the resulting csv file: 
image


backend/routes/download.js
```diff
@@ -34,13 +37,124 @@ const ensureAdminDownloadToken = (req, res, next) => {
...
+              }), {})
+
+              // create update entry objects that contain modified entry data plus:
+              // path, vert and horiz dementions medaiType, provider, videoId
+              const entriesWithTypeData = entries.map(entry => {
+                let entryData = entry.dataValues
+                let entryType = entry.entryType === IMAGE_ENTRY ? 'Image'
+                  : entry.entryType === VIDEO_ENTRY ? 'Video'
+                    : entry.entryType === OTHER_ENTRY ? 'OtherMedia' : ''
+                let newEntry = {
```
chtinahow: Did we not want to include the First and Last name of the student?

frontend/src/Student/components/ShowCard.js
```diff
@@ -136,8 +136,7 @@ const SubmittedEntries = ({ show }) =>
         <EntryThumb entry={entry} />
         {/* If after entry end and before judging end (or if the show is not finalized),
           display "Pending", else display invited or not invited */
-          moment().isBetween(show.entryEnd, show.judgingEnd) ||
-        !show.finalized ? (
+          moment().isBetween(show.entryEnd, show.judgingEnd) || (!show.finalized && moment().isAfter(show.entryEnd)) ? (
```
robmcl4: I think this is more readable as:

```javascript
moment().isAfter(show.entryEnd) && (moment().isBefore(show.judgingEnd) || !show.finalized)
```

I think they're equivalent.

backend/routes/download.js
```diff
@@ -34,13 +37,124 @@ const ensureAdminDownloadToken = (req, res, next) => {
...
+                imageIds.push(entry.entryId)
+              } else if (entry.entryType === VIDEO_ENTRY) {
+                videoIds.push(entry.entryId)
+              } else if (entry.entryType === OTHER_ENTRY) {
+                otherMediaIds.push(entry.entryId)
+              }
+            })
+
+            return Promise.all([Image.findAll({ where: { id: { $in: imageIds } } }),
+              Video.findAll({ where: { id: { $in: videoIds } } }), OtherMedia.findAll({ where: { id: { $in: otherMediaIds } } })
```
robmcl4: I agree with Tina, throw in a `User.findAll(...)` to get all the users so we can include usernames in the csv. Also, groups, so we can include group participants.

backend/routes/download.js
```diff
@@ -34,13 +37,124 @@ const ensureAdminDownloadToken = (req, res, next) => {
...
+                } else if (entry.entryType === OTHER_ENTRY) {
+                  let otherObj = otherMediaIdsToImage[entry.entryId]
+                  return {
+                    ...newEntry,
+                    path: otherObj.path
+                  }
+                }
+              })
+              // Send csv data to browser
+              stringify(entriesWithTypeData, { header: true }, (err, output) => {
```
billdybas: Consider promisifying this so you can do:

```js
stringify()
  .then()
  .catch()
 ```

File: frontend/src/Admin/actions.js [hunk 5]
```diff
@@ -137,3 +137,15 @@ export const downloadZip = showId => (dispatch, getState, client) => {
     '_self'
   )
 }
+
+export const downloadCsv = showId => (dispatch, getState, client) => {
+  const { shared: { auth: { downloadToken } } } = getState()
+
+  // TODO replace this with the deployed url
```
billdybas: After #239 gets merged in, you should make this URL a constant.

frontend/src/Student/components/ShowCard.js
billdybas: I'm a bit confused why this was changed in this PR. What is the change attempting to do?

From my understanding, we should show "Pending" after a student makes a submission. Then, in order for invitations to be visible, judging must be over and the admin must have finalized invitations.

billdybas: Consider moving this into a helper function instead of nested ternaries, as it's becoming confusing to understand at a glance.

kayladavis: Yeah, I just saw it as an issue when I started up. It's a one line fix, so I changed it, thinking, "clean up as you go". Now that I look back though, it's totally a not realistic case (show finalized before entry end), so I'll remove it.

backend/routes/download.js
```diff
@@ -34,13 +37,124 @@ const ensureAdminDownloadToken = (req, res, next) => {
   })
 }
 
+router.route('/csv/:showId')
```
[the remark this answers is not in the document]
kayladavis: Lets handle this an another pr.

kayladavis:
Csv now looks like this: 
image

backend/routes/download.js
```diff
@@ -34,13 +41,298 @@ const ensureAdminDownloadToken = (req, res, next) => {
...
+                      mediaType: '',
+                      videoUrl: ''
+                    }
+                    // Add entry data to data object
+                    if (entry.entryType === IMAGE_ENTRY) {
+                      let imageObj = imageIdsToImage[entry.entryId]
+                      // TODO: change loclahost for deployment
+                      return {
+                        ...newEntry,
+                        path: `//localhost:3000//static/uploads/${imageObj.path}`,
```
robmcl4: Hate to nit here, but let's just start using a config option for this now, since we'd have to do it in a few days anyway. I think just change `backend/.env-sample` to include an all-caps key like uhhhh `STATIC_UPLOAD_BASE_URL`. Be sure to update your `backend/.env` too. And then boom, here we can use
```javascript
`${ config.get('STATIC_UPLOAD_BASE_URL') }/${ imageObj.path }`
```
right here, I think

Also, the `//localhost/...` trick only works in browsers I think (?), the config should probably be `http://localhost/...`

kayladavis:
merging in, as bill'd feedback from the last PR was echoing rob's.

frontend/src/Admin/actions.js
kayladavis: constant made in last commit

PR #254, opened 2018-04-12
Remove a judge's votes when they are unassigned

robmcl4:
adds backend support for deleting a judge's votes if they're removed from a show

TO DO:

- [x] Add modal to front-end warning of this


billdybas:
@robmcl4 I'm going to merge this and mark the caching issue as a known issue.

PR #257, opened 2018-04-16
Download Single Image Button

billdybas:
Also updates some paths

### DIFF
--- a/frontend/src/Admin/components/ShowSubmissionDetails.js
+++ b/frontend/src/Admin/components/ShowSubmissionDetails.js
@@ -153,7 +153,7 @@ const ShowSubmissionDetails = ({ submission }) => (
               href={`${STATIC_PATH}${submission.path}`}
               target='_blank'
             >
-              <Button color='primary'>Download</Button>
+              <Button color='primary'>View Image</Button>
             </a>
           </div>
         ) : null}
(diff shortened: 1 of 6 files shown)


frontend/src/Admin/components/ShowSubmissionDetails.js
```diff
@@ -154,6 +147,16 @@ const ShowSubmissionDetails = ({ submission }) => (
         <dd>{submission.forSale ? 'Yes' : 'No'}</dd>
         <dt>More Copies?</dt>
         <dd>{submission.moreCopies ? 'Yes' : 'No'}</dd>
+        {submission.entryType === PHOTO ? (
+          <div className='text-center'>
+            <a
+              href={`${STATIC_PATH}${submission.path}`}
+              target='_blank'
+            >
+              <Button color='primary'>Download</Button>
```
kayladavis: Change this to 'view image' or something similar, if you're not auto downloading.

billdybas:
To have it auto-download, we'd have to serve the image file w/ [Content-Disposition](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition) header. We could also set the `download` property on the `<a>` tag, but it's not supported in all browsers and only works on the same origin.

I will change the text though to make it more clear what it does.

PR #258, opened 2018-04-17
Frontend Polish

billdybas:
Updates the Visual Design:
- Admin Show Card
- Admin Show Details Tab
- Judge Show Card

### DIFF
--- a/frontend/src/Admin/components/ShowCard.js
+++ b/frontend/src/Admin/components/ShowCard.js
@@ -43,7 +43,7 @@ const ShowCard = props => (
           View Details
         </Button>
         <Button
-          color='warning'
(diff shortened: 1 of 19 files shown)


frontend/src/Admin/components/ShowCard.js
```diff
@@ -20,48 +20,72 @@ const ShowCard = props => (
...
+          style={{ cursor: 'pointer' }}
+          tag={Link}
+          to={`/show/${props.id}/judges`}
+          block
+          outline
+        >
+          View Progress
+        </Button>
+        <Button
+          color='danger'
```
kayladavis: I don't want this to be red. Can we use another color?

PR #266, opened 2018-04-25
Admins can exclude from judging and judges receive only allowed submissions while voting

kayladavis:


backend/resolvers/queries/entryQuery.js
```diff
@@ -32,8 +32,14 @@ export function entries (_, args, req) {
...
     if (!args.showId && !args.studentUsername) {
       return Entry.all()
-    } else if (args.showId) { // Get entries by show
-      return Entry.findAll({ where: { showId: args.showId } })
+    } else if (args.showId) { 
+      // Get entries by show except those not excluded from judging
+      return Entry.findAll({ 
+        where: { 
+          showId: args.showId,
+          excludeFromJudging: false
```
robmcl4: hmm.... is this field nullable? if so, this might be better as "is not true" rather than "is false"

billdybas:
If a submission is excluded from judging, do we remove any votes that submission may have received before it was excluded?

kayladavis:
@bill. I don't think we should remove the votes. Since it's possible to toggle excluded or allowed, I don't think it makes sense to delete the votes. I think instead it makes sense to add a row that shows the allowed or unallowed status, so that when Nanette is inviting people she knows not to allow the excluded ones.

kayladavis:
@billdybas It feels weird for it just to be grey. Any idea about how to show what the grey means?
image

backend/resolvers/queries/entryQuery.js
kayladavis: it's not nullable and it has a default value

PR #271, opened 2018-04-26
Add a button to delete shows

robmcl4:
also does a few other things:

* Adds icons to buttons
* Implements mutation to delete a show
* changes tests to use the schema as created by migrations, not the auto-magic sequelize does

image

image


billdybas:
Because of #221, you'll need to update your icons. We no longer use `react-icons` and use font-awesome directly. You can see examples of the new syntax throughout our codebase.

PR #273, opened 2018-04-26
Update READMEs

billdybas:
More detailed READMEs on how to install and contribute to the backend and frontend, as well as deployment instructions.

TODO:
- [x] Backend README


deploy/deploy.sh
```diff
@@ -168,7 +171,14 @@ rm -r $DEST_FOLDER "$DEST_FOLDER.zip";
 # Finish
 #
 cat << FINMSG
-+----------------------+
-| Finished installing. |
-+----------------------+
+   ______      ____                   ______      __
```
billdybas: The backticks ` need to be changed here, as in bash, they're treated as "execute what's inside them"

PR #279, opened 2018-04-28
Fix Printable Report Title Bug

billdybas:
"show" was not defined until after the app mounted,
so the app would crash when trying to set the page title.

It now uses a tool specifically made for setting the page title.

Note, having `<Fragment>` be a child of `<DocumentTitle>` is intentional. `<DocumentTitle>` expects to only have one child component.

I also added in propTypes for this component.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -55,6 +55,7 @@
     "react-addons-shallow-compare": "^15.6.2",
     "react-apollo": "^2.0.0",
     "react-dates": "^15.0.0",
+    "react-document-title": "^2.0.3",
     "react-dom": "^16.0.0",
     "react-dropzone": "^4.2.3",
     "react-hot-loader": "^3.1.1",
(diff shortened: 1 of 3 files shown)


frontend/package.json
```diff
@@ -55,6 +55,7 @@
     "react-addons-shallow-compare": "^15.6.2",
     "react-apollo": "^2.0.0",
     "react-dates": "^15.0.0",
+    "react-document-title": "^2.0.3",
```
kayladavis: Does this need to be added to the webpack config?

billdybas: Nope. 

This guide explains what I did - [code splitting](https://webpack.js.org/guides/code-splitting/)

By default, our app and all dependencies will be bundled into the same chunk. But by code splitting, it allows different chunks to be loaded in parallel by the browser, causing first render times to be quicker. 

Bundles can also be cached, say if we updated part of our app, but not a dependency, only the main chunk would have to be downloaded again, though I think in our case since we add unique hashes to our chunks, that wouldn't happen.

My process for creating the chunks was looking at larger dependencies that shared similar purposes. So, there's a chunk for React "runtime" dependencies, one for larger React libraries, one for GraphQL libraries, and everything else in the Main chunk. I tried to keep the chunks around the same size, though the Main chunk is larger than the rest.

This dependency is so small, I don't think it needs to go into the Lib chunk, so by default it'll go into the Main chunk.

PR #281, opened 2018-04-30
Code Cleanup & Backend Security Best Practices

billdybas:
- Don't show Sequelize warnings
- Update Sequelize Models to v4
- Removing unused code & Adding comments
- Update some outstanding frontend error handling
- Disable X-Powered-By Header
- Disable CORS in production

The only thing I'm not 100% on is the CORS change. Because we serve the frontend and backend on the same domain, `gallerygateway.rit.edu`, I don't think we need CORS enabled, but I may be wrong. We would need it for local dev because they're run on different ports.

### DIFF
--- a/frontend/src/shared/components/LoginSwitch.js
+++ b/frontend/src/shared/components/LoginSwitch.js
@@ -50,11 +50,9 @@ class LoginSwitch extends Component {
       case 'STUDENT':
         return <StudentPage />
       default:
-        // If for some reason, we receive an unknown user type, render the splash page
-        // TODO: Make an error page that says:
-        // 'Oops! An error occurred. Try clearing your storage and
-        // cookies and refreshing the page. If the problem persists, contact ______.'
-        return <HomePage />
+        // If for some reason, we receive an unknown user type, throw an error,
+        // so that the error boundary can catch it and render the error page
+        throw new Error('Unknown User Type')
     }
   }
 }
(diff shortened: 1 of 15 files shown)


frontend/src/shared/components/LoginSwitch.js
```diff
@@ -50,11 +50,9 @@ class LoginSwitch extends Component {
...
         return <StudentPage />
       default:
-        // If for some reason, we receive an unknown user type, render the splash page
-        // TODO: Make an error page that says:
-        // 'Oops! An error occurred. Try clearing your storage and
-        // cookies and refreshing the page. If the problem persists, contact ______.'
-        return <HomePage />
+        // If for some reason, we receive an unknown user type, throw an error,
+        // so that the error boundary can catch it and render the error page
+        throw new Error('Unknown User Type')
```
billdybas: I ran into weird edge case last night I need to debug - but this error was always being thrown when trying to go to the home page- I think it has to do with having an old or expired JWT

billdybas: I figured out this issue. If you're not signed in, `mapStateToProps` will pass in an `undefined` `user` prop, which causes the default prop for this component to be used, which is `{}`. Our check in `render` is `if (!user)` which is incorrect because an empty object is truthy. The check just needs to be changed to `if (!user.type)`

